PR #8, opened 2012-01-04
File Formats will be supported coffee-script

hisayan:
I usually use the coffeescript.
It is very easy to implement.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -2,6 +2,7 @@
 var Sys = require('sys'),
     File = require('fs'),
     Yaml = require('yaml'),
+    Coffee = require('coffee-script'),
     FileSystem = require('fs');
 
 // Static members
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -8,7 +8,8 @@
   "directories": {"lib": "./lib", "config": "./config", "test": "./test"},
   "dependencies": {
     "yaml" : "0.2.x",
-    "vows" : "0.5.x"
+    "vows" : "0.5.x",
+    "coffee-script" : "~1.2.0"
```
lorenwest: I'm curious - what's the ~ (tilde) syntax mean?

lib/config.js
```diff
@@ -2,6 +2,7 @@
 var Sys = require('sys'),
     File = require('fs'),
     Yaml = require('yaml'),
+    Coffee = require('coffee-script'),
```
lorenwest, line 5: Thank you for adding this format.  The pull request looks good.  

In order to merge the request and publish a new branch it will need to be accompanied with a test.  If you would like to write it, then I'll publish as soon as it's ready.  Otherwise I can write it, but it'll be a week or two before I can get to it.

It's actually very easy.  Create a .coffee configuration and test it in test/2-config-test.js.  You can see how other formats are tested in there.

The current branch is v0.4.9.  I will create it in github so you can attach the pull request to that branch.

Thanks,
-Loren

package.json
hisayan: ref. http://groups.google.com/group/commonjs/browse_thread/thread/a1abe3ffb7203ef4

It can be replace ">=1.2.0" or something.

lib/config.js
hisayan, line 5: Thanks. 

I want to try writing configuration and test in few days. 
and work at current branch. 

my English is not good.

PR #10, opened 2012-01-11
Moved sys to util for node 0.6.x

arifamirani:
Can you please accept this pull request? It gives a fugly warning in 0.6.x

Thanks,
Arif


lib/config.js
```diff
@@ -1,5 +1,5 @@
 // Dependencies
-var Sys = require('sys'),
```
lorenwest, line 2: Thank you for noticing this.

PR #12, opened 2012-01-24
Use 'util' instead of 'sys' to squash 0.6.x warning 

abh:
(it's been called 'util.js' since 0.2.x).


lib/config.js
```diff
@@ -1,5 +1,5 @@
 // Dependencies
-var Sys = require('sys'),
+var Sys = require('util'),
```
lorenwest, line 2: It seems as if I've fixed this about 5 times.  Thanks for the pull request.

Issue #57, opened 2013-04-11
Back to defaults on runtime


josx:
Is there any way to back to defaults values on runtime from code?
I think this would be usefull.

PR #58, opened 2013-04-16
ResetRuntime File and getOriginalConfig

josx:
I have to work on node-config to support new functionality.
The complete work is to use it from webpage (thru backbone/backbone.io [Here](https://github.com/inaes-tic/backbone.io/commit/5523162c60aa326a3e7e4db0185a2c088a73048e))

So I need to rollback runtime changes #57 , and get Original Configuration.

### DIFF
--- a/test/2-config-test.js
+++ b/test/2-config-test.js
@@ -17,6 +17,7 @@ var vows = require('vows');
     newWatchedValue = Math.floor(Math.random() * 100000);
 
 var runtimeJsonFilename = __dirname + '/config/runtime.json';
+var runtimeJsonFilenameBak = runtimeJsonFilename+'.bak';
 
 /**
  * <p>Unit tests for the node-config library.  To run type:</p>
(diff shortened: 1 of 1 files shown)


lib/config.js
```diff
@@ -1134,6 +1134,40 @@ Config.prototype._isObject = function(obj) {
...
+ * <p>
+ * This method allows you to reset the whole runtime.json generated file.
+ * </p>
+ *
+ * @method resetRuntime
+ * @return {boolean} TRUE if can reset file, FALSE if cant
+ */
+Config.prototype.resetRuntime = function() {
+    try {
+        FileSystem.writeFileSync(runtimeJsonFilename, '{}');
```
lorenwest: Synchronous functions are only appropriate at bootstrap time.  Because this can be called after the require('config'), it should be made asynchronous, so it doesn't stop a running server.

test/2-config-test.js
```diff
@@ -257,6 +257,21 @@ exports.ConfigTest = vows.describe('Test suite for node-config').addBatch({
     },
     'Module default values are persisted': function(err, runtimeObj) {
       assert.equal(runtimeObj.TestModule.parm3, 1234);
+    },
+    'The resetRuntime() method is available': function() {
+      assert.isFunction(CONFIG.resetRuntime);
+    },
+    'Runtime Configuration is empty': function() {
+      assert.isTrue(CONFIG.resetRuntime());
```
lorenwest: This doesn't actually test that the runtime configuration is empty (as it says).

test/2-config-test.js
```diff
@@ -257,6 +257,21 @@ exports.ConfigTest = vows.describe('Test suite for node-config').addBatch({
...
+    }
+  },
+
+  'Assuring you can get originalConfig': {
+    topic: function() {
+      return CONFIG.getOriginalConfig();
+    },
+    'The getOriginalConfig() method is available': function() {
+      assert.isFunction(CONFIG.getOriginalConfig);
```
lorenwest, line 288: This tests that the function is present, but it doesn't assure that you can get the originalConfig as the test implies.

lib/config.js
josx: I didnt notice about it. Thanks!
Changes in commit   7467d39

test/2-config-test.js
josx: Here, please check 8f3d21b

test/2-config-test.js
josx, line 288: Any clue on how to test _CONFIG.getOriginalConfig_ against _orginalConfig_ ?
I cant access to _originalConfig_ static members from tests.

PR #61, opened 2013-04-27
Json-schema Validation

josx:
In [Caspa](https://github.com/inaes-tic/mbc-caspa) we need [Json-schema](http://json-schema.org/) for validate config file and to have info about config fields.

I have added jayschema module to validate.

If you want to merge it, please tell me so I can add documentation and tests.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -16,9 +16,11 @@ var DEFAULT_CLONE_DEPTH = 6,
     CONFIG_DIR = process.env['NODE_CONFIG_DIR'] || process.cwd() + '/config',
     runtimeJsonFilename = process.env['NODE_CONFIG_RUNTIME_JSON'] || CONFIG_DIR + '/runtime.json',
     appInstance = process.env['NODE_APP_INSTANCE'],
+    schemaValidatorFilename = process.env['NODE_CONFIG_SCHEMA_VALIDATOR'] || CONFIG_DIR + '/schema_validator.json',
     originalConfig = null,       // Not including the runtime.json values
     runtimeJson = {},            // Current runtimeJson extensions
     runtimeJsonWatcher = null,   // Filesystem watcher for runtime.json
+    schemaValidator = {},
     isQueuedForPersistence = false;
 
 /**
@@ -103,6 +105,7 @@ var Config = function() {
   var t = this;
   t._loadFileConfigs();
   t._persistConfigsOnChange();
+  t._schemaValidation();
 };
 
 /**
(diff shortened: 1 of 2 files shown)


toolbear:
>  to where you don't control the configuration files

Actually, there's a quite common use-case even with small projects and with controlled/trusted configurations: having the app fast-fail when a configuration hasn't been set. Unless you're disciplined about reviewing every commit message after doing `git pull`, this can occur every time a collaborator defines a new, required configuration. It's a big time saver when `npm start` fails immediately with a useful message versus the app failing later when attempting to use a feature, both in development and in production and production-like environments.

Prior art: [webonise/configise](https://github.com/webonise/configise) is a related configuration framework that includes a config verification step via a bespoke API.

lorenwest:
Are you volunteering to write this component?

toolbear:
Not at the moment. I came upon this while searching for whether anyone was using JSON schema in this way. I thought sharing another perspective would be useful, but if it's counterproductive let me know and I'll delete my comments to reduce the noise.

lorenwest:
It's not counter productive at all.  I think this would be helpful, and it's good to keep this thread going in case someone wants to step up and add JSON schema support.

gajus:
> Are you volunteering to write this component?

Are you upon to a PR that would add JSON schema support?

OnkelTem:
@lorenwest 

> You can validate the config files before calling node-config, and contain the json-schema dependency in your local project.

I didn't quite understand how it can be achieved. 

The list of config file candidates is maintained and generated by `node-config`. But even using `.util.getConfigSources()` wouldn't suffice as we end up with a list of `.parsed` yet unmerged data structures. So we should first merge them and only them validate against schema, right?

One would use `config.get()` (without parameters) to get the final merged config to validate it against a schema, but... there seems to be no way to get this config: get requires precisely one argument, a key.

Please could you clarify this?

OnkelTem:
Answering to myself. Probably I was looking for `config.util.toObject()`.

markstos:
> I didn't quite understand how it can be achieved.

One option is to use a separate process. For example, use [CUE Lang](https://cuelang.org/) to validate a structure and possibly produce a JSON file as a result. `node-config` would then read that file.

OnkelTem:
> > I didn't quite understand how it can be achieved.
> 
> One option is to use a separate process. For example, use [CUE Lang](https://cuelang.org/) to validate a structure and possibly produce a JSON file as a result. `node-config` would then read that file.

That's not the problem which had to be addressed. `node-config` reads many input file formats from different locations, merging them and providing the final result for the program. It doesn't look a good idea for me to interject into this config reading process. What has to be check is the final result.

PR #76, opened 2013-08-06
Added flag for disabling the write of runtime.json

Osterjour:
For performance issues we don't want to write config changes in a file. So we need a flag to disable this behaviour.

### DIFF
--- a/doc/template/overview.tmpl
+++ b/doc/template/overview.tmpl
@@ -71,7 +71,10 @@
   <p>
     In the running application, each configuration parameter is internally
     monitored for changes, and persisted to runtime.json upon change.
-    This keeps configurations in sync in a mutli-node deployment.
+    This keeps configurations in sync in a mutli-node deployment. If you want to
+    disable this persistance and sync you can set the environment variable
+    <i>NODE_CONFIG_PERSIST_ON_CHANGE</i> to "N". This will lead to lower I/O if you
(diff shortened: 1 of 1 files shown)


lib/config.js
```diff
@@ -102,7 +102,7 @@ var DEFAULT_CLONE_DEPTH = 20,
 var Config = function() {
   var t = this;
   t._loadFileConfigs();
-  t._persistConfigsOnChange();
+  if(process.env['DONT_WRITE_CONFIG_ON_CHANGE'] != "Y") t._persistConfigsOnChange();
```
lorenwest: This is a good feature to add.  A few requests:

1) Change the environment variable name to match other node config environment variable names, and state it in the positive:  NODE_CONFIG_PERSIST_ON_CHANGE="Y/N" (with the default being "Y")

2) Make the change inside _persistConfigsOnChange(), because it should disable persistence regardless of where it's called (there is another place _persistConfigsOnChange() is called).

3) Document the environment variable usage and default (probably in doc/template/overview.tmpl)

4) Extra bonus points:  Write a test to verify it works, and continues to work through future changes.

My suggestion is to add something like this to the top of _persistConfigsOnChange():
if (process.env['NODE_CONFIG_PERSIST_ON_CHANGE'] === 'N') {
    return;
}

Issue #110, opened 2014-02-27
Using config.watch only runs one time


sshrefler:
var config = require('config');

config.watch(config, null, function(object, propertyName, priorValue, newValue) {
  console.log("change");
});

I would expect that when I keep making changes to runtime.json, it would keep calling the function.  It only happens once.

PR #111, opened 2014-02-27
Change FileSystem.watch to FileSystem.watchFile

sshrefler:
This fixes https://github.com/lorenwest/node-config/issues/110.   watchFile will cause the runtime.json to be checked for changes more than once on osx.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -503,7 +503,7 @@ Config.prototype.watchForConfigFileChanges = function(interval) {
   try {
     if (FileSystem.watch) {
       // This is the latest
-      runtimeJsonWatcher = FileSystem.watch(RUNTIME_JSON_FILENAME, {persistent:false}, function(event, filename) {
+      runtimeJsonWatcher = FileSystem.watchFile(RUNTIME_JSON_FILENAME, {persistent:false}, function(event, filename) {
 
         // Re-attach watcher on inode change (happens when some editors save a file)
         // Wait for the O/S rename to complete, then re-watch the file


lib/config.js
```diff
@@ -503,7 +503,7 @@ Config.prototype.watchForConfigFileChanges = function(interval) {
   try {
     if (FileSystem.watch) {
       // This is the latest
-      runtimeJsonWatcher = FileSystem.watch(RUNTIME_JSON_FILENAME, {persistent:false}, function(event, filename) {
```
lorenwest, line 506: If you look at the if that gets us into this area, you'll notice it runs this because watch is available (vs. watchfile)

Watchfile is the old way, watch is the new way.  Both are in node-config for backward compatibility, but at some point watchfile will go away.

I think the issue is deeper than changing watch to watchfile.  I think the issue is that the inode representing runtime.json is being changed becuase your editor writes files a certain way (by moving the old file and creating a new file).

lorenwest, line 506: Here's some conversation about the issue: https://groups.google.com/forum/#!topic/nodejs/Pz1TkZ3wyRY

sshrefler, line 506: Ah, that makes sense.  Any suggestions on a more robust way to watch runtime.json for a file change?  On OSX watchFile actually worked as advertised but that seems to be flaky in other environments

lorenwest, line 506: How are you changing runtime.json?  I've found that if I do it in vi, it breaks the watch because vi doesn't overwrite - it moves the old file and creates a new file, then removes the moved file.

When you overwrite runtime.json it works as expected.  An alternative could be to put a heartbeat timeout watching for runtime.json having a different inode, but I think this is an edge case when hand editing runtime.json vs. anything happening out in the data centers - such as a node.js or python script overwriting it.

sshrefler, line 506: Thanks for tracking that down!  I was actually using Sublime on OSX.   If i simply change runtime.json using nano it overwrites correctly.

Thank you!

PR #135, opened 2014-08-19
update support for Coffee-script >=1.7.0

th507:
Starting from Coffee-script 1.7.0, the API for requiring `.coffee` in Node.js has changed as described in https://github.com/jashkenas/coffeescript/issues/3349#issuecomment-34015414 .

This patch updates the API for parsing config with the latest version of Coffee-script.

Tests run fine.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -777,9 +777,12 @@ util.parseFile = function(fullFilename) {
     }
     else if (extension == 'coffee') {
       // Lazy load the coffee-script extension
-      Coffee = Coffee || require('coffee-script');
-      if (Coffee.register) {
-        Coffee.register();
+      if (!Coffee) {
+        Coffee = require('coffee-script');
+        // Handle coffee-script > 1.7.0
+        if (Coffee.register) {
+          Coffee.register();
+        }
       }
       // Use the built-in parser for .coffee files with coffee-script
       configObject = require(fullFilename);


lib/config.js
```diff
@@ -777,7 +777,7 @@ util.parseFile = function(fullFilename) {
     }
     else if (extension == 'coffee') {
       // Lazy load the coffee-script extension
-      Coffee = Coffee || require('coffee-script');
+      Coffee = Coffee || require('coffee-script/register');
```
lorenwest: This may break earlier installs of coffee-script.  The devDependencies are just for node-config testing, but lib/config.js is for everyone.  According that link, it looks like you can call a register function.  Would something like this work for both versions of coffee-script?

```
Coffee = Coffee || require('coffee-script');
if (Coffee.register) {
  Coffee.register();
}
```

lib/config.js
```diff
@@ -778,6 +778,9 @@ util.parseFile = function(fullFilename) {
     else if (extension == 'coffee') {
       // Lazy load the coffee-script extension
       Coffee = Coffee || require('coffee-script');
+      if (Coffee.register) {
+        Coffee.register();
+      }
```
lorenwest, line 786: This would call `register()` each time a file with `.coffee` extension is loaded. I don't know if that would cause any problems, but this seems better:

```
// Lazy load the coffee-script extension
if (!Coffee) {
  Coffee = require('coffee-script');
  if (Coffee.register) {
    Coffee.register();
  }
}
```

PR #141, opened 2014-09-29
Use lodash for deep cloning.

### LINKED ISSUE #139: Regular expressions are converted into {}
That's about all I know. I have a property that looks like

```
{
  prop: [ /regexp/ ]
}
```

and it's being converted into

```
{
  prop: [ {} ]
}
```

diversario:
This way Date, Regexp and other non-plain objects
are preserved while copying.

Fixes #139.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -884,7 +884,7 @@ util.attachProtoDeep = function(toObject, depth) {
  * subject to the following conditions: The above copyright notice and this permission
  * notice shall be included in all copies or substantial portions of the Software.
  */
-util.cloneDeep = function cloneDeep(parent, circular, depth, prototype) {
+util.cloneDeep = function cloneDeep(parent, depth, circular, prototype) {
   // maintain two arrays for circular references, where corresponding parents
   // and children have the same index
   var allParents = [];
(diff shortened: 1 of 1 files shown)


lib/config.js
```diff
@@ -869,37 +870,82 @@ util.attachProtoDeep = function(toObject, depth) {
...
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the “Software”), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+ * of the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions: The above copyright notice and this permission
+ * notice shall be included in all copies or substantial portions of the Software.
  */
-util.cloneDeep = function(obj, depth) {
+util.cloneDeep = function cloneDeep(parent, circular, depth, prototype) {
```
lorenwest: The [util.cloneDeep()](https://github.com/lorenwest/node-config/wiki/Using-Config-Utilities#clonedeepcopyfrom-depth) API cannot change without introducing a backward incompatibility.

According to the [Future Compatibility](https://github.com/lorenwest/node-config/wiki/Future-Compatibility) spec, a 1 year waiting period would be required for backward incompatible changes.

diversario: Yep, I missed that. I can switch the arguments so `depth` is the second one.

PR #178, opened 2014-11-23
Initial implementation of Strict mode.

markstos:
- Off by default. Set process.env.NODE_CONFIG_STRICT_MODE to a true value
  to enable.
  - When on, there must be an explicit config file matching NODE_ENV if
    NODE_ENV is set.
  - When on, there must be an explicit config file matching NODE_APP_INSTANCE
    if NODE_APP_INSTANCE is set
  - When on, NODE_ENV must not match 'default' or 'local' to avoid ambiguity.

If a strictness check fails, an exception will be throws at require-time.

If this is excepted, I would update the wiki notes to note that the feature is
experimental and the strictness rules are subject to change.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -116,10 +116,8 @@ var Config = function() {
   util.extendDeep(t, util.loadFileConfigs());
   util.attachProtoDeep(t);
 
-  // If requested, perform strictness checks and possibly throw an exception.
-  if (process.env.NODE_CONFIG_STRICT_MODE) {
-    util.checkStrictly(t);
-  }
+  // Perform strictness checks and possibly throw an exception.
+  util.runStrictnessChecks(t);
 };
 
 /**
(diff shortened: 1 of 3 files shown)


lib/config.js
```diff
@@ -114,6 +115,11 @@ var Config = function() {
   // Merge configurations into this
   util.extendDeep(t, util.loadFileConfigs());
   util.attachProtoDeep(t);
+
+  // If requested, perform strictness checks and possibly throw an exception.
+  if (process.env.NODE_CONFIG_STRICT_MODE) {
+    util.checkStrictly(t);
+  }
```
lorenwest: These are useful tests even if strict mode isn't specified.  I would suggest having util.checkStrictly() return an error string.  If a string were returned it could throw an error if strict mode was selected (as it does now), or issue a WARNING if strict mode isn't selected (like other warnings).

lib/config.js
```diff
@@ -1470,6 +1476,43 @@ util.getRegExpFlags = function (re) {
...
+    return Path.basename(src.name);
+  });
+
+  // Throw an exception if there's no explicit config file for NODE_ENV
+  var anyFilesMatchEnv = sourceFilenames.some(function (filename) {
+      return filename.match(NODE_ENV);
+  });
+  if (NODE_ENV && !anyFilesMatchEnv) {
+    throw new Error("FATAL configuration problem. NODE_ENV value of '"+NODE_ENV+"' did match any deployment config file names.");
+  }
```
lorenwest, line 1484: There's a pattern I'd like us to continue, whereas error/warning strings emitted contain a link to further explanation on the node-config wiki.

PR #192, opened 2015-01-07
Added TOML support

jasonhansel:


lorenwest:
Hi @jasonhansel - thanks for the pull request.  It looks great!

Would you mind adding a section on TOML in the [configuration files](https://github.com/lorenwest/node-config/wiki/Configuration-Files) wiki page?  It follows a consistent pattern for all formats, so it should be easy.  Anything special about TOML files that you'd like to add (such as why it's better than sliced bread), feel free.

Cheers,
-Loren

jasonhansel:
Done! Thanks for merging this!

lorenwest:
Published 1.10.0 to NPM, and you made it onto the contributors list on the front page!

PR #193, opened 2015-01-07
Hjson Support

laktak:
I've added support for Hjson, a configuration file format that's similar to JSON but with a simplified syntax.

It supports #, // and /**/ style comments as well as avoiding trailing/missing comma and other mistakes. For details and syntax see http://laktak.github.io/hjson/

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -12,6 +12,7 @@ var Yaml = null,    // External libraries are lazy-loaded
     PPARSER = null,
     JSON5 = null,
     TOML = null,
+    HJSON = null,
     deferConfig = require('../defer').deferConfig,
     DeferredConfig = require('../defer').DeferredConfig,
     Utils = require('util'),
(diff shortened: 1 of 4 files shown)


lorenwest:
Hi @laktak,

Would you mind adding a section on Hjson in the [configuration files](https://github.com/lorenwest/node-config/wiki/Configuration-Files) wiki page? It follows a consistent pattern for all formats, so it should be easy.

Thanks for adding this feature,
-Loren

markstos:
@laktak HJSON looks very similar to [JSON5](http://json5.org/). What are the key differences? JSON5 seems worth a mention from the HJSON website, as they seem likely to be confused or compared.

laktak:
@lorenwest thanks, I've updated the wiki.

@markstos hjson doesn't need quotes or commas and supports multiline strings with proper indentation. I had a section linking to other formats but it didn't fit with the site. I might add it again as a wiki.

lorenwest:
Just published v1.11 with hjson support - thanks for adding this feature!

PR #198, opened 2015-02-17
Date type doesn't overwrite

### LINKED ISSUE #197: Date type doesn't overwrite
I am having problems with Date type not being overwritten. I have prepared a repo to show the error.

https://github.com/axelhzf/node-config-error

The problem is a have a default.js

```js
module.exports = {
  str: "A",
  date: new Date(2015, 0, 1)
};
```

and a development.js

```js
module.exports = {
  str: "B",
  date: new Date(2015, 0, 2)
};
```

index.js

```js
var config = require("config");
console.log(config.get("str")); // B
console.log(config.get("date")); // Thu Jan 01 2015 00:00:00 GMT+0000 (WET) it should be Thu Jan 02 2015 00:00:00 GMT+0000 (WET)
```

The date property has the default value instead of development value

axelhzf:
fixes #197

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1311,7 +1311,9 @@ util.extendDeep = function(mergeInto) {
 
       // Extend recursively if both elements are objects and target is not really a deferred function
       var isDeferredFunc = mergeInto[prop] instanceof DeferredConfig;
-      if (util.isObject(mergeInto[prop]) && util.isObject(mergeFrom[prop]) && !isDeferredFunc) {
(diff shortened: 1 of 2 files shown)


lorenwest:
This looks like a healthy pull request - would you mind adding a test that fails in the current version, and passes after this change?

axelhzf:
I have already added the test. Any other comment on the PR?

lorenwest:
Nope - that looks good.  Thanks!

PR #205, opened 2015-04-09
Support DeferredConfig-like objects

fracmak:
So the scenario I'm trying to support is very much an edgecase, but hear me out. I have a modularized project, where the root project is sucking in code from various sub node_modules, which include their own config settings. The problem I have is that each project has it's own copy of node-config, so when I suck in their config, all DeferredConfig objects don't match the `instanceof DeferredConfig` check in the project and they don't get resolved. This just allows non-object objects with a resolve function to be treated like a DeferredConfig object. 

Thoughts?

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -723,6 +723,10 @@ util.loadFileConfigs = function() {
   return config;
 };
 
+function isDeferredConfig(value) {
+  return value instanceof DeferredConfig || (typeof value == 'object' && typeof value.resolve == 'function');
+}
+
 // Using basic recursion pattern, find all the deferred values and resolve them.
 util.resolveDeferredConfigs = function (config) {
     var completeConfig = config;
@@ -737,8 +741,8 @@ util.resolveDeferredConfigs = function (config) {
                       _iterate(prop[property][i]);
                   }
               } else {
-                  if (prop[property] instanceof DeferredConfig ) {
-                    prop[property]= prop[property].resolve.call(completeConfig,completeConfig)
+                  if (isDeferredConfig(prop[property])) {
+                    prop[property]= prop[property].resolve.call(completeConfig,completeConfig);
                   }
                   else {
                     // Nothing to do. Keep the property how it is.
--- a/test/3-config/default.js
+++ b/test/3-config/default.js
@@ -8,6 +8,11 @@ var config = {
 
 };
 
+function DeferLike() {}
+DeferLike.prototype.resolve = function() {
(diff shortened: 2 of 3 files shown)


lorenwest:
This changes an underlying philosophy that configurations aren't overriden by sub-modules.  Node-config supports the concept of sub-modules declaring default configurations, while allowing the application using the sub-module to override those defaults.

When configuring complex systems, it's important that configuration values (other than defaults) are managed in one place.  I have a history of projects where configurations are stored in an indeterminate number of places across the filesystem, and node-config specifically does not support that functionality.

fracmak:
Ya, I can completely understand the problems with sub module configurations and playing a game of "who's overriding my value?!?" But you hinted at the problem I've run into. Default values can live in sub modules which is what I'm trying to accomplish. I've been able to get everything working except for this DeferredConfig problem. I can technically live without it, but it would be nice to loosen the requirements of the matching exactly the DeferredConfigs and support objects that are resolvable (ie, has a resolve function). Do you have any real concern with expanding that feature?

lorenwest:
We have a [sub-module defaulting mechanism](https://github.com/lorenwest/node-config/wiki/Sub-Module-Configuration), which if used for defaults should be sufficient.  

The concept of dynamic is at opposition with the concept of default.

Maybe @markstos has an opinion on the use of DeferredConfig for specifying sub-module defaults.

fracmak:
That's what I'm trying to fix, if the default config values are from a node_module that does a require('config/defer').deferConfig, it's a different config instance from the parent module which then fails the check for resolving the deferred config.

markstos:
I lean towards a wait-and-see approach to see if this comes again. The "deferred config" feature is already advanced feature, and this adds a bit more complexity to it. 

I'm comfortable with the idea of the patch that would allow a "duck typing" style check. ("If it's an object that quacks with with a "resolve" method, then it must be a duck"), I'm just not sure about endorsing/recommending this mixing of deferred configuration and sub-modules.

If we continue to hear this feature request from multiple users, that could sway my opinion.

lib/config.js
```diff
@@ -723,6 +723,10 @@ util.loadFileConfigs = function() {
   return config;
 };
 
+function isDeferredConfig(value) {
+  return value instanceof DeferredConfig || (typeof value == 'object' && typeof value.resolve == 'function');
```
Klortho, line 727: A little bit stronger criterion would be better here, I think. What about `... || value.constructor.name === 'DeferredConfig'` ?

PR #209, opened 2015-04-17
Used Path.Join to concantenate folder location strings. This fix will fix issues using the project in windows

michaelpuzon:
...x issues with using the project in windows

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -624,16 +624,16 @@ util.loadFileConfigs = function() {
 
   // Initialize parameters from command line, environment, or default
   NODE_ENV = util.initParam('NODE_ENV', 'development');
-  CONFIG_DIR = util.initParam('NODE_CONFIG_DIR', process.cwd() + '/config');
+  CONFIG_DIR = util.initParam('NODE_CONFIG_DIR', Path.join( process.cwd(), 'config') );
   if (CONFIG_DIR.indexOf('.') === 0) {
-    CONFIG_DIR = process.cwd() + '/' + CONFIG_DIR;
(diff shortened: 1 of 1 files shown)


lorenwest:
The underlying `require('fs')` code does windows slash translation for us, so this won't result in any behavior change.

On one hand, this suggests that we've thought about windows slashes.  On the other hand, it shows that we think it matters, when it doesn't.

tomato tamato

michaelpuzon:
Hi guys, I encountered this problem because I am developing a node app in windows setting. It was not loading the config folder at all, and the hard-coded slash  on the string concantenation was the culprit.

markstos:
If this fixes things for a Windows user and doesn't break things for other users, I'm fine with the patch. I know the pattern is common in other languages.. at least Perl uses the same kind of pattern to deal with slash portability.

lorenwest:
Now I'm confused.  Node-config has been working under windows for years, and your experience differs from my understanding of the fs package.

I wonder if the fs package really doesn't do slash translation.  Anyone with a windows machine that would be able to try it out?

```
var fs=require('fs');
var content = fs.readFileSync('/somedir/somefile.txt');
```

I think that will work on windows as long as there's a \somedir\somefile.txt file out there. It's more for general understanding of this issue than for this pull request.  I'm on board with this pull request.

-Loren

markstos:
The path-translation feature of `fs` sounded familiar to me, but I don't see it documented in the official docs:

https://nodejs.org/api/fs.html

What version of Node.js on Windows did the `fs` slash translation fail on @michaelpuzon ?

michaelpuzon:
Hi. I am on v0.12.0 for node.js running on 64-bit windows 7.

PR #214, opened 2015-05-05
updated cson dependency

### LINKED ISSUE #212: cson dependency not up to date
After installing both `config` and `cson`, parsing cson doesn't work out of the box because `parseSync` doesn't exist anymore in the cson package (v3.0.1).

It looks like `config` is dependant on an older version of cson (v1.6.1), which is why it still works in tests.

dsimidzija:
Let's see if this closes #212 :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,6 @@ language: node_js
 before_install:
   - npm update -g npm
 node_js:
+  - "0.12"
   - "0.10"
   - "0.8"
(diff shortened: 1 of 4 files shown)


lorenwest:
Thank you for your work on this!

lorenwest:
Just published to NPM at 1.13.0

dsimidzija:
Wow, that was fast..thanks, glad I could help :)

PR #216, opened 2015-05-09
Handle the presence of a Unicode BOM character in any config file by …

johndkane:
…stripping it from the start of the file.

Add /test/7-* tests into the package to test Unicode situations with, specifically the BOM stripping functionality included in this update.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -801,6 +801,7 @@ util.parseFile = function(fullFilename) {
   // Try loading the file.
   try {
     fileContent = FileSystem.readFileSync(fullFilename, 'UTF-8');
+    fileContent = fileContent.replace(/^\uFEFF/, '');
   }
   catch (e2) {
     throw new Error('Config file ' + fullFilename + ' cannot be read');
--- a/test/7-config/defaultNoBOM.json
+++ b/test/7-config/defaultNoBOM.json
@@ -1,0 +1,3 @@
+{
+  "siteTitle" : "A valid site title from default.json"
+}
--- a/test/7-config/defaultWithUnicodeBOM.json
+++ b/test/7-config/defaultWithUnicodeBOM.json
@@ -1,0 +1,12 @@
+﻿{
+  "Customers": {
+    "dbName":"from_default_json",
+    "dbPassword":"password will be overwritten.",
+    "dbPassword2":"password will be overwritten.",
+    "lang":["en","es"]
+  },
+  "AnotherModule": {
+    "parm1":"value1"
+  },
+  "staticArray": [2,1,3]
+}
--- a/test/7-config/test-7.json
+++ b/test/7-config/test-7.json
@@ -1,0 +1,12 @@
+{
+  "Customers": {
+    "dbName":"from_default_json",
+    "dbPassword":"password will be overwritten.",
+    "dbPassword2":"password will be overwritten.",
+    "lang":["en","es"]
+  },
+  "AnotherModule": {
+    "parm1":"value1"
+  },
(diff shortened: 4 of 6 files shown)


lorenwest:
The code looks straight forward, but somehow it's failing the integration tests: https://travis-ci.org/lorenwest/node-config/builds/61839306

It's probably an issue with the tests vs. the actual code, but the tests have to pass before the code can be merged.  Do the errors indicate what might be happening?

johndkane:
I'll try to give the tests another go this weekend. They were built on a windows box and I used 'node test file.js' to run it; maybe my process is off... Thanks for the heads up.

johndkane:
Turns out I don't have a good understanding of testing with node. It might take me a bit of time to get up to speed but I see the effort as worthwhile.  

In the meantime, if anybody can help me understand why the test file 7-unicode-situations.js that I created is trying to find its configuration in the 6-config/ folder instead of 7-config/ folder it would be much appreciated. 

~John

markstos:
Try your calls to set process.env in the scope of the topic function
instead of globally.

On May 11, 2015 10:30 PM, "John K" <notifications@github.com> wrote:
>
> Turns out I don't have a good understanding of testing with node. It
might take me a bit of time to get up to speed but I see the effort as
worthwhile.
>
> In the meantime, if anybody can help me understand why the test file
7-unicode-situations.js that I created is trying to find its configuration
in the 6-config/ folder instead of 7-config/ folder it would be much
appreciated.

johndkane:
Oops, the test file is still bad. It's a case of 'it works on my machine' but I'm on this. Eventually the test file will work and I'll understand how to more proficiently handle project updates in the future from my end. Please bear with the newb!

markstos:
@johndkane  Here's a bit more detail about what I think went wrong with your test:

All the tests are loaded into one process. Several are trying to the same global variables, so order and timing matter.  First, all the files get loaded, and all the "global" functions that are outside functions get run immediately, perhaps not in a predictable order. So, this is *not* where you want to set a global variable that you only want to affect your own test. By changing a global value inside of a test function the change should happen *just before* the code you are testing needs it. Because tests are run sequentially and not in parallel, the test should run in the content of the expected global value. 

I've run into this issue myself before. 

    Mark

johndkane:
Thank you @markstos  for the support!  It's much appreciated. The Unicode BOM tests are now green-lighting on my trial version of TravisCI here https://travis-ci.org/johndkane/node-config

I think it's ready for prime time. If not then I'll head back to the woodshed with vows testing.

johndkane:
Question: me being new to this, would you rather I fork and make a new pull request having one clean commit, or can you merge just the latest changes I made without acquiring the history of troubleshooting cruft incurred while debugging through uploads to TravisCI?  

I see the history from my forked repo is being relayed above and I don't want to pollute the project if it's preventable. Let me know what is preferable. 

Thank you.

markstos:
@johndkane A simple, clean commit history is preferred. In your branch you can use `git rebase -i` to squash many commits into one and reword some commit messages. When you are done rebasing, you can `git push -f ... ...` to "force push" your changes to the branch your pull request was made from. Then this page will be updated, reflecting that the contents of the pull request have changed. We can then merge the clean commit history. 

Thanks.

johndkane:
@markstos , thank you once again.  I think it's good now.  That was the most painful one line code fix I've made to date but the education was well worth it.

markstos:
You are welcome. I have been there myself.

lorenwest:
Thanks for the addition!

PR #213, opened 2015-05-05
node-config/#211 Enable environment variables to be parsed in a specified format

leachiM2k:

PR #219, opened 2015-05-18
Refactor: split up parseFile into a second function parseString for string-only based formats

leachiM2k:
As mentioned in pull-request https://github.com/lorenwest/node-config/pull/213 this pull request refactors parseFile and extracts string-based format parsing to a second function. Only .js, .coffee and .iced remains in parseFile.
This is needed for a later use.


lorenwest:
Thanks for splitting this out.  It's going to make the other commit easier to review.

PR #243, opened 2015-09-02
fix attachProtoDeep for setModuleDefaults

benkroeger:
setModuleDefaults wouldn't apply the config prototype functions (like `.get()`) to the modules config in the case that no other configuration was provided for this module.  
Also resetting the `checkMutability` flag according to initParam `ALLOW_CONFIG_MUTATIONS`, since the setModuleDefaults function could have been called after the first config.get and then would expose the defaults mutable.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -380,11 +380,11 @@ util.watch = function(object, property, handler, depth) {
  * @param defaultProperties {object} - The default module configuration.
  * @return moduleConfig {object} - The module level configuration object.
  */
-util.setModuleDefaults = function(moduleName, defaultProperties) {
+util.setModuleDefaults = function (moduleName, defaultProperties) {
 
   // Copy the properties into a new object
   var t = this,
-      moduleConfig = util.cloneDeep(defaultProperties);
+    moduleConfig = util.cloneDeep(defaultProperties);
 
   // Set module defaults into the first sources element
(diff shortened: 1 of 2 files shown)


markstos:
Thanks for the contribution. Could you also contribute a test suite updates that illustrates the bug?

benkroeger:
test case added

markstos:
Looks OK to me, but I don't use this feature. What do you think @lorenwest ? 

I noticed that setModuleDefaults is only documented on this one wiki page that I found:
https://github.com/lorenwest/node-config/wiki/Sub-Module-Configuration

I was surprised to see that it wasn't on the "internal utilities" page as well.  But for a relatively-hidden feature, there are 13 issues that reference it:

https://github.com/lorenwest/node-config/search?q=setModuleDefaults&type=Issues&utf8=%E2%9C%93

I would want to be sure that fixing this case doesn't re-open one of the other related cases.

lorenwest:
This looks like a healthy change.  Thanks for the test - it helped to understand the issue.

lorenwest:
@markstos - As to this possibly regressing, this PR looks solid but it's a risk for any change.  There are 164 tests to help reduce the risk, and if it does regress that flushes out the need for the 165th or 166th.

It's a calculated risk, offset by the benefit of progress.

lorenwest:
Published to npm today at version 1.16

PR #252, opened 2015-10-12
Accept options from init

klinquist:
Option priority:
a) Command line
b) Environment variable
c) Initialization options (new)
d) default values


lorenwest:
I believe this PR is doomed because node-config doesn't accept construction parameters.  It doesn't create independent copies for each module that uses node-config.  The nature of configurations is that they're app-global, so anyone that calls `require('config')` gets the same configurations. If we allowed each module to specify different parameters, then we wouldn't be fulfilling the goal of node-config.

Fortunately there's a pattern for making this work, and it only needs to be run before the first call to node-config (generally in the app bootstrap).

klinquist:
Understood. Thanks for the comments!

PR #258, opened 2015-11-11
Correctly handle parsing of inline JSON

elliotttf:
This arose becasue of the way comments are stripped from JSON. As comments will remain supported in node-config (#256) this change will correctly handle parsing JSON strings regardless of whether or not a comment is encountered.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -947,8 +947,7 @@ util.parseString = function (content, format) {
     }
   }
   else if (format === 'json') {
-    // Allow comments in JSON files
-    configObject = JSON.parse(util.stripComments(content));
+    configObject = JSON.parse(util.stripJSONComments(content));
   }
   else if (format === 'json5') {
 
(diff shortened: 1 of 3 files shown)


markstos:
I reviewed the diff and it looks-good-to-me.

lib/config.js
```diff
@@ -1434,9 +1449,11 @@ util.stripYamlComments = function(fileStr) {
  * @protected
  * @method stripComments
  * @param fileString {string} The string to strip comments from
+ * @param stringRegex {RegExp} Optional regular expression to match strings
```
lorenwest: Could you provide a better explanation?  Is this regex supposed to match comments, non-comments, full lines, partial lines?

lib/config.js
```diff
@@ -1423,6 +1423,21 @@ util.stripYamlComments = function(fileStr) {
...
+ * This method will call util.stripComments but has a simplified regex that
+ * will work with JSON strings.
+ *
+ * @protected
+ * @method stripJSONComments
+ * @param fileString {string} The string to strip comments from
+ * @return {string} The string with comments stripped.
+ */
+util.stripJSONComments = function(fileStr) {
+  return util.stripComments(fileStr, /("[^"]*")/g);
```
lorenwest, line 1439: I'm not a regex guru.  Could you add a comment here describing this regex?  Something like

```
// Match everything starting with a quote up to and including a matching quote.
```

It appears that this would miss a lot - such as /\* ... */ comments, but I can't tell without a good description of the intention of this regex.

elliotttf, line 1439: Sure, just to clarify, this is effectively equivalent to the regex used by `util.stripComments` but is targeted for valid JSON. The one in `util.stripComments` can begin with either `'` or `"` and will capture everything to a closing `'` or `"`. Since this method is only concerned with JSON we don't need to worry about `'` and can focus only on strings that begin and end with `"`. The regex is used  to create valid JSON and remove comments, so skipping over strings within comments is completely acceptable. The test was updated to confirm that valid JSON is generated and I've successfully tested locally with a few different combinations of strings.

lorenwest:
Merged and published today with 1.17.0

PR #260, opened 2015-11-17
test with a more complex quote

adityabansod:


lorenwest:
Sweet - thanks!  I'm going to merge this in, knowing it fails the Travis CI, in preparation for either a rollback or bugfix soon (as in today).

adityabansod:
:+1:

markstos:
I vote for rolling back the JSON change until a better fix is available.

lorenwest:
If @elliotttf or someone appears soon with a fix to this, that would be a little better than a rollback.  I'm thinking about doing a rollback in a couple hours if this isn't resolved by then.

elliotttf:
That's a bummer, I'm looking into this, but as I mentioned in #256 anything we do with regexes is liable to be brittle.

elliotttf:
#261 should resolve this.

PR #261, opened 2015-11-17
Fix nested quotes with support for empty strings

### LINKED ISSUE #259: Escaped Quotes in JSON break
Hi there --

This recent PR (https://github.com/lorenwest/node-config/pull/258) caused escaped quotes inside of a JSON value to break parsing.

e.g. 

```json
"patientApptOffer": "<a href=\"http://localhost:3000/offers/reply?id={{system.contact.value}}\">Test String</a>",
```

gets parsed to 
```json
"patientApptOffer": "<a href=\"http:
```

which then blows up when `JSON.parse` gets its hands on it.

elliotttf:
So it turns out the problem in #258 wasn't so much inline JSON as JSON with empty strings. This _should_ fix those problems, but again all regex based solutions are going to have corner cases.

Fixes #259

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1436,7 +1436,7 @@ util.stripYamlComments = function(fileStr) {
 util.stripJSONComments = function(fileStr) {
   // Because we are stripping comments from JSON we are only concerned with
   // strings that begin and end with a double quote.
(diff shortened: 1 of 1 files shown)


lorenwest:
Published in 1.17.1

PR #263, opened 2015-11-19
Added support of NODE_DEFAULT_CONFIG_FILENAME.

teologov:
Hi! I've added ability to change default config filename. 
This variable allows to set new filename, if it's required.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -638,6 +638,7 @@ util.loadFileConfigs = function() {
   APP_INSTANCE = util.initParam('NODE_APP_INSTANCE');
   HOST = util.initParam('HOST');
   HOSTNAME = util.initParam('HOSTNAME');
+  DEFAULT_CONFIG_FILENAME = util.initParam('NODE_DEFAULT_CONFIG_FILENAME', 'default');
 
   // This is for backward compatibility
   RUNTIME_JSON_FILENAME = util.initParam('NODE_CONFIG_RUNTIME_JSON', Path.join(CONFIG_DIR , 'runtime.json') );
@@ -659,7 +660,7 @@ util.loadFileConfigs = function() {
   }
 
   // Read each file in turn
-  var baseNames = ['default', NODE_ENV];
+  var baseNames = [DEFAULT_CONFIG_FILENAME, NODE_ENV];
 
   // #236: Also add full hostname when they are different.
   if ( hostName ) {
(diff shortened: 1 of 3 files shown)


markstos:
@teologov, if there's a conflict in `/config`, why not use NODE_CONFIG_DIR instead?

lorenwest:
Could you describe the problem that this is intended to solve? There may be another way of solving it that doesn't add to the complexity of our pre-config configurations.

teologov:
@lorenwest I am integrating node-config into a really big project. 
We've already got a big config.properties file. 
I don't want everyone involved into project to rename existing config filename. 

@markstos Could you please explain, what do you mean? 
Thanks.

lorenwest:
For local development, each machine can have its own configuration file, extended from the default.

I'm having a hard time understanding the need for multiple default files, or how they would organize them if this pull request were accepted.  The purpose of a default file is that it's the one base for all extensions to extend.  If it's getting too big, then pointing to a different one isn't going to resolve the size problem.

Help me understand your large project - is it large in number of developers, number of modules, number of deployment types, number of applications using a common configuration, or ?

Do you understand the multiple file merging capability described in the wiki page here? https://github.com/lorenwest/node-config/wiki/Configuration-Files

teologov:
Well, the thing is, that I don't need multiple default files. But I would like to set the name of the default file.

The project is large in both directions. There are a lot of developers, and the project is really large.

I would like to migrate this project to use node-config module. We've already got a default config file, but filename differs from "default.EXT". 

With that said, I would like to keep on old approach with naming of the default config file in my project. Otherwise, everyone have to change it's config filename.

lorenwest:
node-config is designed to handle large enterprise installations with hundreds of servers and many large teams of developers.  It is unlikely that your use case is much different than the tens of thousands of other people using node-config.

So I'd really like to understand if, you really did find an edge case that requires a change to node-config, or if you are simply not understanding how it works.  

Having many ways to accomplish the same task is not a virtue we aspire to when accepting pull requests.

So, please - other than "it's a large project", help me understand your issue.  If your situation truly is one in a million requiring an edge-case solution, I'd be happy to include this functionality.

teologov:
> If your situation truly is one in a million requiring an edge-case solution

No, it's not.

Well, I still can use the current approach.
I do understand how it works. This is just a small improvement, and nothing more.

If this won't be included - I am fine with it.

Thanks for the great module!

markstos:
@teologov Is the project using git? Could you rename the default file using "git mv" or symlink the old file file name to `config/default.js` ? Git supports storing symlinks.

teologov:
@markstos Yes, this is not a problem. Thanks.

PR #267, opened 2015-11-20
Resolve defered values in predictable order for consistent results. Fixes #265

### LINKED ISSUE #265: Bug: order of resolution of deferred configurations is arbitrary, potentially leading to inconsistent configurations
Consider a config file in which one deferred config references another:

```javascript
// using defer functions is optional. See example and docs below.
 var defer = require('config/defer').deferConfig

module.exports = {
  a : "value a",
  b : defer(function () {
    return this.a;
  }),
  c : defer(function () {
    return this.b;
  }),
};
```

The resulting configuration might or might not work because the order in which the defer'ed functions is arbitrary. ([We use the `for (property in object)` pattern](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in)).

If `b` is resolved before `c`, you get a working configuration. If `c` is resolved before `c` (which is what happened when I tried today,  then b will still point to a "deferred" function at the time of resolution instead of a value, and the configuration will be broken.

Perhaps the "arbitrary" order resolution is generally static, but it could vary between JavaScript interpreters, or versions of Node.js. 

A solution could be to throw in a "sort"  as we iterate objects, to create a predictable object traversal order. That would also allow developers a way explicitly make sure one value is resolved before it is referred to.

markstos:
Before, we used the `for..in..` construction to iterate over object
properties, which returns the keys in an arbitrary order which is
usually-but-not-always the key insertion order.

Now object keys are traversed in the order returned by calling sort() on
the key names

Having a predictable resolution order could be useful if you want one
defered value to refer to another defered value. Now one should reliably
resolve before the other.

The change passes all the regression tests. No new test coverage was added. Due to the nature, it was unclear how to test the case of the resolution order being unpredictable in different JavaScript environments.

It's possible that the change could cause someone's code somewhere to break, if they were depending on the old, arbitrary resolution order, which happened to be reliable for them.  However, it could be argued it was a bug in their code to depend on the resolution order, when no resolution order was documented.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -747,28 +747,39 @@ util.loadFileConfigs = function() {
 
 // Using basic recursion pattern, find all the deferred values and resolve them.
 util.resolveDeferredConfigs = function (config) {
-    var completeConfig = config;
+  var completeConfig = config;
 
-    function _iterate (prop) {
-      for (var property in prop) {
-          if (prop.hasOwnProperty(property) && prop[property] != null) {
-              if (prop[property].constructor == Object) {
-                  _iterate(prop[property]);
-              } else if (prop[property].constructor == Array) {
-                  for (var i = 0; i < prop[property].length; i++) {
-                      _iterate(prop[property][i]);
-                  }
-              } else {
-                  if (prop[property] instanceof DeferredConfig ) {
-                    prop[property]= prop[property].resolve.call(completeConfig,completeConfig)
-                  }
-                  else {
-                    // Nothing to do. Keep the property how it is.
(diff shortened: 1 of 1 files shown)


lorenwest:
Some people don't like .js over .json because code in configuration can be nondeterministic. You not only have code in your configs, you have deferreds in your code in your configs.  And deferreds in your code in your configs that rely on the order of other deferreds in your code in our configs.

Do you skydive? :)

All kidding aside, I have no problem with this pull request.

PR #271, opened 2015-12-02
Allow specifying a default value to get().

ikogan:
In some cases, having to test with `has()` before a get is quite cumbersome. If you're going to specify a default value if `has()` returns false, it would be easier to simply provide it to `get()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,13 @@ The following examples are in JSON format, but configurations can be in other [f
       ...
     }
 
-`config.get()` will throw an exception for undefined keys to help catch typos and missing values.
+**Default values:**
+
+    var config = require('config');
+    ...
+    app.listen(config.get('server.port', 9000));
+
+`config.get()` will throw an exception for undefined keys to help catch typos and missing values unless a second, `defaultValue` argument is specified, in which case, the default value will be returned instead.
 Use `config.has()` to test if a configuration value is defined.
 
 **Start your app server:**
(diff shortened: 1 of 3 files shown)


lorenwest:
I'm going to have to decline this pull request because it results in configuration values stored in unpredictable locations.

There is one consistent place to store default configurations.  Everyone that uses node-config knows immediately where to find default configurations.  This change would disrupt that consistency.

You may fork node-config if you like this feature, however I suggest re-considering where you store default values. My experience is on teams larger than 1, a consistent location for all configurations is appreciated.

ikogan:
I would agree, except in cases where the host and port are unknown to you until runtime. Cloud9 determines the port and host of your application at runtime and stores it in the environment, you don't have a way of knowing what it'll be until you run the application:

```
Important: use process.env.PORT as the port and process.env.IP as the host in your scripts!
```

If you listen on any other port, it won't be able to route requests to it.

ikogan:
I suppose, though, in those cases you're using `config` in the absence of the environment, so this doesn't really solve that issue.

lorenwest:
There are ways of overriding default configurations in [environment variables](https://github.com/lorenwest/node-config/wiki/Environment-Variables) and on the [command line](https://github.com/lorenwest/node-config/wiki/Command-Line-Overrides).

Both of these represent consistent places of overriding defaults.  Your pull request isn't about overriding defaults, it's about defining defaults.  There should be one place that defaults are defined, and consistent places for overriding these defaults.

Node-config is about having deterministic locations for configurations.  At 3am when you don't know why your servers are acting the way they are, it's valuable to have a consistent place to find your configurations.

ikogan:
Ah, I wasn't aware of custom environment variables. Thanks for educating me. Too much coffee too fast apparently ;-).

PR #280, opened 2016-01-07
 NODE_APP_INSTANCE changed to NODE_CONFIG_APP_INSTANCE to avoid conflicts with PM2

manoloaterol:
PM2 uses the env variable NODE_APP_INSTANCE to manage instances in cluster mode, so projects that uses node-config will fail trying to load default-0.json....default-n.json files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .project
 node_modules
+.idea/
--- a/lib/config.js
+++ b/lib/config.js
@@ -613,7 +613,7 @@ util.getConfigSources = function() {
  * </p>
  *
  * <p>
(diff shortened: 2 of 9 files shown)


markstos:
This change would break backwards compatibility, which we strive to minimize:
https://github.com/lorenwest/node-config/wiki/Future-Compatibility

lorenwest:
haha - strive to minimize is an understatement.  We won't merge anything that breaks backwards compatibility.  

In this case I understand there's another module that came into existence after node-config, and it's a popular module that happened to choose the same variable name. 

I wouldn't mind supporting both variables as a migration to NODE_CONFIG_APP_INSTANCE, but a simple replacement won't do.

PR #289, opened 2016-02-05
Handle null subobject case

wmertens:
With Yaml, an empty subobject becomes `null`, and `typeof null` is `object` which makes getImpl descend into it.

Test fails without config.js change and passes with.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -145,7 +145,8 @@ var getImpl= function(object, property) {
   if (elems.length <= 1) {
     return value;
   }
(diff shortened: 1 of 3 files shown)


lorenwest:
This looks like a good commit - thanks for contributing!

markstos:
Does this affect the distinction that we make that "undefined != null" ?

I presume if there was a problem with this update, it would break some existing test.

wmertens:
@markstos this only changes the case where you are trying to dereference a null value.

Issue #225, opened 2015-06-13
Read submodule defaults from their own config directory (PR welcome)


lorenwest:
Proposal to add 1-arg behavior to `config.setModuleDefaults(moduleName, defaults)`.  If the 2nd argument isn't specified, it loads defaults from the /config directory within the module specified by `moduleName`.

This serves many purposes:

```
* It moves submodule configurations out of code
* It puts submodule configuration into a well known place
* It provides a common place to put configuration documentation
* It allows running the submodule as an app, with overrides for tests, etc.
* It prevents namespace clash by utilizing the submodule package name
* It doesn't break backwards compatibility
* The 2-arg mechanism can be deprecated, with ample warning
```

When this change is made, change the wiki page to document only this behavior.  Make a note about the features the 2-arg behavior doesn't have, and that it will be going away in the future (at least 1 year).

Establish a convention for submodules that they be configured by passing overrides to the module constructor, or by overrides in the app /config directory.  This allows the submodule to be used by apps that also use `node-config` as well as apps that use other configuration systems.

lorenwest:
This will still have the problem outlined by issue #201 so maybe some progress on that issue can be made at the same time.

lorenwest:
The new `setModuleDefaults()` will need to specify the module, since `node-config` isn't able to detect the module directory of the code calling `setModuleDefaults()`.  

Existing method:

    setModuleDefaults(moduleName, configObject);

Proposed alternative:

    setModuleDefaults(module);

If a module is passed as the first arg, the module name will be retrieved from the module's package.json file, and the default config will be loaded from the /config/default.{type} file under the module directory.

Bruce17:
I forked your module and implemented a working fix to this issue (I needed a solution asap).

Have a look at it: https://github.com/Bruce17/node-config/commit/9a6b19c5df601580dd2aafa354c45b7723626d1e

I implemented a new method called `initConfig` which accepts a directory. After loading the module `config`, I call this method to re-init the config module with a additional directory.

```js
// Allow the configs to be overwritten
process.env.ALLOW_CONFIG_MUTATIONS = true;

// Force the module "node-config" to load this module's config files
var config = require('config');
config.initConfig(__dirname + '/../config/');
```

This could be a way to solve this issue. I'm currently using this solution in a working project.

lorenwest:
[...]

PR #292, opened 2016-02-12
Read submodule defaults from their own config directory

AndrewMRoberts:
This pull request is in reference to https://github.com/lorenwest/node-config/issues/225.

This fix allows for submodules to use a single arg version of `util.setModuleDefaults()`. If utilized, config will load in the submodules configuration values from the submodules `/config` directory and name them based on the name of the submodule pulled from it's `package.json` to avoid naming conflicts.

As a result, this led to also changing the `getImpl()` function to first grab configuration values from a submodule object if it exists, and if the submodule object does not exist, just uses the top level configuration object as normal.

Values can still be overridden from the top level `/config` directory as normal. It also allows for submodules to be run as an individual apps for testing purposes, since at that point it will recognize it's own config directory as the top level config directory.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -483,9 +483,9 @@ util.setModuleDefaults = function (moduleName, defaultProperties) {
   // we are not making t[moduleName] immutable immediately,
   // since there might be more modifications before the first config.get
   if (!util.initParam('ALLOW_CONFIG_MUTATIONS', false)) {
-    checkMutability = true;
+    checkMutability = true; 
   }
-  
+
   // Attach handlers & watchers onto the module config object
   return util.attachProtoDeep(t[moduleName]);
 };
@@ -1767,5 +1767,5 @@ var config = module.exports = new Config();
 var showWarnings = !(util.initParam('SUPPRESS_NO_CONFIG_WARNING'));
 if (showWarnings && Object.keys(config).length === 0) {
   console.error('WARNING: No configurations found in configuration directory:' +CONFIG_DIR);
-  console.error('WARNING: To disable this warning set SUPRESS_NO_CONFIG_WARNING in the environment.');
+  console.error('WARNING: To disable this warning set SUPPRESS_NO_CONFIG_WARNING in the environment.');
 }
--- a/lib/config.js
+++ b/lib/config.js
@@ -145,7 +145,7 @@ var getImpl= function(object, property) {
       value;
 
   // If this is being called from a submodule, we have to modify the object being accessed
-  var value = (getCallingFilePath() !== process.cwd() && object[packageName]) ? object[packageName][name] : object[name];
+  var value = (callingFileDirectoryPath() !== process.cwd() && object[packageName]) ? object[packageName][name] : object[name];
   if (elems.length <= 1) {
     return value;
   }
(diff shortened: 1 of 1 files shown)


luislobo:
+1

lorenwest:
I must have fallen asleep when reviewing this. It appears to be a valuable feature.

It's a fairly substantial addition - I'd like to merge it but am hesitant without some tests that show how it's intended to be used, and that prevent regression. Anyone up for writing some tests?

lib/config.js
```diff
@@ -138,15 +138,21 @@ var util = Config.prototype.util = {};
  * @return value {mixed} - Property value, including undefined if not defined.
  */
 var getImpl= function(object, property) {
+  // Check for null to acquiesce tests for checking has() on a null object
+  if (object === null) return undefined;
```
lorenwest, line 142: Does this have anything to do with submodule /config dirs? If not I'd prefer it be in a separate change list, with a separate test that outlines the problem.

lorenwest:
On further review I can't merge this code - it's too aggressive. `utils.setModuleDefaults()` is for setting defaults, not for setting complete overrides. If you had a /config directory in a submodule, any overrides in that config directory would be there for running from the submodule, not from the containing module.

A sub module can't know who contains it, nor the environment names the containing module uses. It shouldn't be allowed to apply an override to, say, a `production` environment in hopes that the containing application will be run with a `NODE_ENV=production`. All `production` overrides should be in a single file in the containing app. Having overrides scattered around the filesystem is a data center nightmare.

Also, the `getImpl()` changes introduce heavyweight synchronous file system operations on every get. Synchronous code can be run at application startup, but never during a long running application.

If the only thing this did was to allow module *defaults* to be set from a /config/defaults.{js} file within a submodule vs a hardcoded object, and only when `utils.setModuleDefaults()` was called, this would be a healthy addition.

Here's an easier way of loading sub module defaults from a file:

in the containing app `default.js`:

```
module.exports = {
  ...
  mySubModule: require('my-sub-module/config/default.js).mySubModule,
  ...
}
```

Robotriceratops:
The fact of the matter was that the problem I was trying to solve was attempting to let all submodules know what environment they are running in and having them override defaults that the containing module may have overridden. The environment I'm using the library under is very heavy on dependent submodules my team has written and without the ability to override these config values, I was getting collision on commonly used config value names all over the place.

Also because the submodules were depended upon by multiple different modules, declaring the overrides for the submodule in the containing module's config was the exact scenario you described, a data center nightmare, and had me copying the exact same values in many different files for all of our modules, all under different repositories. Keeping the values in the singular submodules directory and repository was a much simpler and elegant solution.

For the same reason, we didn't want our submodules to be running with production variables when running the main modules in test when doing integration testing. The only solution to this was to again copy all of our test overrides for the submodules into each containing module, which quickly exacerbated our copying problems even more.

However, I'm fine with closing the pull request if this is not behavior that is desired. I agree and understand that it was a rather aggressive change.

lorenwest:
If the containing app and contained module are tightly coupled, you can do this without modification to node-config.  For example, each containing app `config/production.js` can do this:

```javascript
module.exports = {
  ...
  mySubModule: require('my-sub-module/config/production.js').mySubModule,
  ...
}
```

This pattern is available without modifications to node-config. Does that fulfill your needs?

jdmarshall:
I think there's a much simpler PR wanting to break out of this one, but there are a couple of changes necessary to loadFileConfigs, which is really not set up to handle multiple calls (and definitely not recursive ones).

jdmarshall:
For instance, if util.setModuleDefaults(myName, configDir) -> util.setModuleDefaults(myName, util.loadFileConfigs(configDir)[myName]);

markstos:
Because this PR has conflicts and hasn't seen much interest in 5 years, I'm closing it. Someone is welcome to read through the history here  for an updated PR which addresses the concerns and conflicts.

PR #295, opened 2016-02-17
Suggested WIP implementation

### LINKED ISSUE #294: Use directories instead of dashes
Hi,

I just wondered if it would make sense to enhance the lib to support directories if an `index.js` is in there (a simple `require` in the end) I already saw #234 but I think thats a bit different. 

Btw I wouldn't mind to implement it on my own.
```
default.EXT => require('./default')
default-{instance}.EXT => require('./default/{instance}')
{deployment}.EXT => require('./{deployment}')
{deployment}-{instance}.EXT => require('./{deployment}/{instance}')
```

The background for me is the following, we have currently 4 environments and 10 instances (e.g. countries) which results in a huge load of files in one directory and I would like to organize it in a folder structure and hide a bit the complexity :smile:.

Currently it looks a bit like this

```
default.js
default-DE.js
default-FR.js
...
staging.js
staging-DE.js
staging-...
live.js
live-...
```

And 2 more. Maybe I could drop one of the deployments and use more default, but still our config differs by country and having 40 files in one folder seems a bit too much. The order here is quite clear, we could use the same logic as before.

evangelion1204:
This is a suggested WIP implementation of the described feature and open for discussion :smile: Just hacked the logic sort of in. Maybe small refactoring would be helpful.

Closes #294

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -677,12 +677,12 @@ util.loadFileConfigs = function() {
 
   baseNames.push('local', 'local-' + NODE_ENV);
 
-  var extNames = ['js', 'json', 'json5', 'hjson', 'toml', 'coffee', 'iced', 'yaml', 'yml', 'cson', 'properties'];
+  var extNames = ['js', 'json', 'json5', 'hjson', 'toml', 'coffee', 'iced', 'yaml', 'yml', 'cson', 'properties', ''];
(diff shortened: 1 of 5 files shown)


markstos:
Closed because #294 is closed.

PR #304, opened 2016-04-01
Feature to search parent config path

### LINKED ISSUE #303: Submodule config loading
I have two express applications, they both contains `config` like this

```js
--App A
----config
----server
------index.js // require('App B index.js')
--App B
----config
----server
------index.js  // gets config from App A :(
```
If I try to include `App B`.`index.js` inside `App A`.`index.js` , `App B` code will load `config` from `App A`.

Is it possible for each application to still use their own config folder rather than inheriting from parent ?

sushantdhiman:
Closes #303 

This PR will allow `node-config` to recursively check for parent `config` folder. This will allow sub module to define their own configuration inside their own child `config` folder.

This behavior can be overridden by `NODE_CONFIG_DIR`, so `node-config` will always prefer to `NODE_CONFIG_DIR`, if not set it will lookup for parent `config` folder chain, if no config folder found it will default back to `process.cwd() + '/config'` folder. 

cc @lorenwest

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -410,9 +410,9 @@ util.setModuleDefaults = function (moduleName, defaultProperties) {
   // we are not making t[moduleName] immutable immediately,
   // since there might be more modifications before the first config.get
   if (!util.initParam('ALLOW_CONFIG_MUTATIONS', false)) {
-    checkMutability = true; 
+    checkMutability = true;
   }
-  
+
   // Attach handlers & watchers onto the module config object
   return util.attachProtoDeep(t[moduleName]);
 };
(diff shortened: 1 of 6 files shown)


sushantdhiman:
@lorenwest Please review when you get chance :)

lorenwest:
Hi @sushantdhiman,

I like the concept of submodule configs, and have written about an approach in #225. 

This code appears to have the effect of discovering an alternate config directory by looking through the filesystem based on the location of the specific node-config install.

Unfortunately this is too radical to be included. Configs often contain security information, and their locations must be well known, not discovered. They can only be at the top of the application directory, not buried within a config directory of the first of many node-config installs in a deep node_module tree.

At least this could introduce an incompatibility where a new install brings in different configs, but at worse this opens up a security hole where bad-guy submodule gets to load their configs instead of the configs in the application directory.

You are welcome to publish this as a fork of node-config as it serves a specific purpose, but I cannot merge this into node-config.

-Loren

sushantdhiman:
Thanks for your review @lorenwest, I understand that discovering parent config will make module too flexible for its own good.

I will try to work on your ideas from #225 when I get chance.

Issue #305, opened 2016-04-08
Migration to JS standard


tusharmath:
Hi,
[StandardJS](http://standardjs.com/) is a pretty decent tool for avoiding all the bikeshedding related to linting JS projects.

I observed that there is no `.eslintrc` in the project nor are there any guidelines as to what is being currently followed. It would be great if we could just simply migrate to standard because adding changes right now creates way too many linting errors on my IDE (webstorm) and I am sure it happens with other devs as well.

Thanks.

markstos:
This is the first time this has come up. You are welcome to create a PR that passes all the StandardJS checks so we can see what the differences look like.

tusharmath:
Alright, this is a lot of work, let me give it a shot.

vroudge:
Hello,

Just dropping by to say that StandardJS is not the standard at all as of now. The JS community has outbursted a lot against the owner of the repo which tries to imply that some choices on the JS coding style he's done in his repo are NOT standard at all. Wink wink semicolons.

One should refer to https://github.com/airbnb/javascript, for instance, or any other styleguide and its linter, that are consider much more knowlegeable in the industry.

Thanks for config, by the way, love it.

tusharmath:
@vroudge I can not disagree with you here. feross/Standard apparently is more popular and its easy to get started with it because of the variety of tools that are already available. I think a better solution would be to add a `.eslintrc` file instead for this project.

vroudge:
Airbnb-styleguide, for instance, is muuuuuuch more popular than Standard. No possible doubt, just check the npm packages for ESLint downloads/month, or just the stars on the github repo. :+1:

tusharmath:
@vroudge I had not seen the stars/forks on github. Nevertheless, you have solid point and that's why I created https://github.com/lorenwest/node-config/pull/310.

PR #307, opened 2016-04-08
Standard JS integration

tusharmath:
Fixes : #305


markstos:
@tusharmath this pull request doesn't pass our Travis tests because there are hundreds of StandardJS failures.

tusharmath:
@markstos That's the intent. You know when to merge, when all the issues are fixed.

tusharmath:
Total Errors: `1935`

tusharmath:
Total Errors: `232`

tusharmath:
@markstos Take a look at the errors. Apart from style errors, it has also highlighted errors such as — using *undeclared variables* , using *==* instead of *===*. This could lead to potential issues in production.

tusharmath:
@markstos There is one particular failure —

```bash
/node-config/lib/config.js
  247:40  error  'fnName' is not defined
```
I am not sure how to fix it. May be you can help me here?

lorenwest:
That's a straight-up bug.  Instead of

```
...config.' + fnName + '()...
```

It should be

```
...config.watch()...
```

It looks like that code was copied from a more generic place into the watch function, without making it specific to watch().

lorenwest:
This PR has too many changes.  Is there a way to change a configuration such that the errors are significantly reduced to match the coding style of the package, and the only errors are real improvements like ```==``` to ```===``` ?

tusharmath:
@lorenwest No we can't configure it. Check the [FAQ](http://standardjs.com/index.html#i-disagree-with-rule-x-can-you-change-it).

If the changes are too big to review, I would recommend you go commit by commit. That way u will be able to review all the changes.

tusharmath:
@lorenwest Some weird issue on node `0.8` can you try re-running after clearing cache on travis?

tusharmath:
Looks like the issue was with `snazzy` module which doesn't work with version `0.8` because of its dependency on `inherits` module. I have removed it for now as it just a prettfier for standardjs's output. Once the issue is fixed, I guess we can re-integrate it later.

PS: Build is fixed!

lorenwest:
This was a large amount of work and as much as I dislike the size of the PR, the point is to endure some pain in order to achieve consistency.

My concern is that manual changes are intermixed with automated changes, making it impractical to detect the changes due to reformatting vs. the changes due to manual code. Normally that'd be mitigated by tests, but the tests were changed along with the code.

My suggestion to getting this merged is to split the work into multiple pull requests:

1) All manual changes needed
2) Automated changes to lib/config.js alone, showing existing tests pass
3) Automated changes to the rest of the package, including tests
4) Travis hooks so earlier commits don't break travis

I sincerely appreciate the work you've put into this and the consistency it brings to the library. If you can break it into these multiple commits, I will take the time necessary to get them committed and pushed to NPM once they're all in.

-Loren

tusharmath:
Closing this infavour of https://github.com/lorenwest/node-config/pull/310

PR #312, opened 2016-04-22
improve highlight

vinicius73:


lorenwest:
Such a simple change, great visual value.  Thanks!

PR #315, opened 2016-04-27
Remove deprecated __lookupGetter__ use

### LINKED ISSUE #313: TypeError: mergeFrom.__lookupGetter__ is not a function (PR welcome)
Node v6 removed the non-standard `__lookupGetter__` method from `Object.prototype`. This causes an application with a default and environment specific config to immediately crash.

`Object.getOwnPropertyDescriptor(mergeFrom, prop).get` is the standard way to access the getter, and `Object.defineProperty(mergeInto, prop, Object.getOwnPropertyDescriptor(mergeFrom, prop))` would properly copy it over.

thetalecrafter:
fixes #313.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -410,9 +410,9 @@ util.setModuleDefaults = function (moduleName, defaultProperties) {
   // we are not making t[moduleName] immutable immediately,
   // since there might be more modifications before the first config.get
   if (!util.initParam('ALLOW_CONFIG_MUTATIONS', false)) {
-    checkMutability = true; 
+    checkMutability = true;
   }
-  
+
   // Attach handlers & watchers onto the module config object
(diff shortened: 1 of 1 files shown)


thetalecrafter:
While I tested locally on 4.4.3, the versions in travis threw an exception. I will update to fix.

markstos:
I see from Travis that this passes on  Node 0.8, Node .10 and Node 0.12.

@lorenwest can Travis be updated to also test the Node 4.4, the current LTS releases, as well as Node 6?

lorenwest:
Yes, it's in .travis.yml and that doesn't need an npm push. New commit to any branch will kick off the changes.

markstos:
@lorenwest I've updated .travis.yml now to also test 4.4 and 6.0. Thanks for the reminder on how to configure that.

PR #316, opened 2016-04-27
Handle node v6 SyntaxError message containing position data

pwwolf:
Fix to catch the JSON parse exceptions which have changed in Node v6.
Position info has been appended to the end of the message, so it now looks like "Unexpected token / in JSON at position X"


markstos:
@lorenwest I've also updated package.json and History.md so everything is set to publish a new release for v6 compatibility.

PR #318, opened 2016-05-09
Fixes #266 - implement recursive resolver for deferreds, using getters

### LINKED ISSUE #266: Bug: defer'ed values can end up in the final configuration object depending resolution order. AKA "the double-defer bug"
This is closely related to #265.

Consider the following configuration, where one deferred configuration value refers to another configuration value:

``` javascript
// using defer functions is optional. See example and docs below.
 var defer = require('config/defer').deferConfig


module.exports = {
  a : "value a",
  b : defer(function () {
    return this.a;
  }),
  c : defer(function () {
    return this.b;
  }),
};
```

Regardless of whether #265 is fixed, it's possible that a deferred value will be referenced in another deferred value.   

This happened to us in today in a complex configuration which wasn't easy to spot right away and where it's desirable to have these values to deferred, to eliminate configuration repetition.

The current result is a kind of a "silent failure" where the the deferred function becomes the value of one of the functions, which later becomes a bug in the application. 

Some possible improvements would be:
1. Fail explicitly. Throw an exception noting that nested defer'ed values have been encountered and aren't supported.
2. Resolve-on-demand. When an unresolved is encountered in a defer function, we would resolve it immediately. 

Resolve-on-demand sounds attractive, but has challenges. Right now raw objects are used in the defer functions. We use an explicit method in the functions to access config properties, which would involve changing the syntax of in defer functions (at least if you want this safety check). 

Other possibility might be to modify the 'this' available to defer functions using a custom `get`  key for the properties. When properties were attempted to be read, we could resolve-on-demand.  This custom 'getters' would not be needed on the final object. 

Failing explicitly in this case is simpler.

Klortho:
Here is the improvement to the deferred resolver logic that we discussed on #266. 

@lorenwest, I completely understand your hesitancy about adding new code, as you described it, "to make it appear to be improved", but let me say a couple of things.

First, as I mentioned before, this feature has nothing whatsoever to do with promises, other than the (unfortunate) using the same terms. There is nothing async going on at all, and no "non-determinism". You asked earlier what would happen if someone read one of the config values before it was resolved, and I was confused at the time. But now I see that you were thinking that the config tree is exposed first and that the resolution happens asynchronously at some later time. In fact, the whole resolution process is synchronous, and very fast -- it's little more than a tree walk. No cloning or copying of trees, either -- the results of the "deferred" functions are grafted in place (just like they are now).

Second, this PR is an implementation, almost exactly, of what @markstos  described in [this comment](https://github.com/lorenwest/node-config/issues/266#issuecomment-207520735). One thing I didn't do was to add a boolean to record whether or not any deferreds are present. It's a good idea. I doubt it would be necessary for performance reasons, but if it is, it would be easy to add.

I put the new code into the defer.js module. I spent a long time painstakingly going over the algorithm, to make sure it is robust. I documented it in extensive comments in that module, and also in lib/resolver.md. The algorithm is a bit tricky, but it works really well. With this fix, you can basically write anything you want in terms of config variables that depend on other config variables -- in any order, using objects or arrays, and nested data structures, nested deferreds, etc. 

I wrote a lot of tests, and my working copy passes all the existing tests, but I'm not sure I covered everything. I didn't test it with the submodules feature. I tested a little bit with using class instances instead of plain-old-json objects as config data. There might be issues with deferreds that return those class instances -- but I don't know whether or not these are handled flawlessly with the existing scheme. If you point me to some test cases, I'd be happy to add more tests.

### DIFF
--- a/defer.js
+++ b/defer.js
@@ -1,14 +1,122 @@
-// Create a deferredConfig prototype so that we can check for it when reviewing the configs later.
-function DeferredConfig () {
-}
-DeferredConfig.prototype.resolve = function (config) {};
+// See resolver.md for some notes about how this works.
 
-// Accept a function that we'll use to resolve this value later and return a 'deferred' configuration value to resolve it later.
-function deferConfig (func) {
+// Class that wraps special functions that are evaluated after all config 
+// overrides are in place.
+function DeferredConfig () {}
+
+// Users use this function in their JavaScript config files to specify config 
+// values that depend on others. This returns a DeferredConfig object that
+// wraps the function (refered to simply as "the deferred").
+function deferConfig(func) {
   var obj = Object.create(DeferredConfig.prototype);
   obj.resolve = func;
   return obj;
 }
 
-module.exports.deferConfig = deferConfig;
-module.exports.DeferredConfig = DeferredConfig;
+// The main function that resolves the entire config tree, evaluating and
+// replacing all deferreds.
+function resolve(mainConfig) {
+  var main = new Resolver(mainConfig);
+  main.resolve();
+}
+
+// Resolver objects define getters for each of the properties in a config node.
+// The purpose of the Resolver objects is to proxy those nodes, so that they
+// can participate in expressions inside deferred functions.
+// The constructor itself does not recurse; it only defines getters for the 
+// immediate children.
+var Resolver = function(config, main) {
+  var self = this;
+  if (!main) main = self;
+
+  // The enumerable keys of config that we care about
+  var childKeys = Object.keys(config).filter(function(k) {
+    return config.hasOwnProperty(k) && typeof config[k] !== 'undefined';
+  });
+
+  // Create a non-enumerable property, __data__, to hold bookkeeping data
+  var data = {
+    main: main,
+    config: config,
+    childKeys: childKeys,
+  };
+  Object.defineProperty(self, '__data__', {
+    __proto__: null,
+    value: data,
+  });
+
+  // Make the getters.
+  // Contract: for every enumerable property (k, v) of a config node, the getter
+  // returns either the original atom from the config, or a resolver (never an object or a
+  // deferred)
+  var values = {};  // memoize the results
+  childKeys.forEach(function(key) {
+    Object.defineProperty(self, key, {
+      __proto__: null,
+      enumerable: true,
+      configurable: false,
+      get: function() {
+        if (key in values) return values[key];
+        return values[key] = self.newNode(config[key]);
+      }
+    });
+  });
+};
+
+// resolver.newNode() - passes through or creates resolver tree nodes 
+// corresponding to config tree nodes. The `node` argument can be of any type. 
+// This evaluates deferred functions, and wraps config objects in resolvers, as 
+// needed. Contract: the return value is guaranteed to be an atom or a resolver
+Resolver.prototype.newNode = function(node) {
+  var self = this,
+      main = self.__data__.main;
+  var t = nodeType(node);
+  return (t === 'atom') ? node :
+         (t === 'resolver') ? node :
+         // deferreds are evaluated recursively:
+         (t === 'deferred') ? self.newNode(node.resolve.call(main, main)) :
+         new Resolver(node, main);
+};
+
+// resolver.resolve() - recursively resolves all of the data in the config tree.
+// Contract: after this executes, the `config` node corresponding to this 
+// resolver will be an atom or an object tree such that:
+// - there are no deferreds or resolvers anywhere in the tree
+// - every node in the tree is the corresponding original atom or object from 
+//   the config tree (i.e., there are no clones)
+Resolver.prototype.resolve = function() {
+  var self = this,
+      data = self.__data__,
+      config = data.config,
+      childKeys = data.childKeys;
+  childKeys.forEach(function(key) {
+    var v = self[key];      // either an atom or a resolver
+    if (nodeType(v) === 'atom')
+      config[key] = v;
+    else {  // resolver
+      v.resolve();
+      config[key] = v.__data__.config;
+    }
+  });
+};
+
+// There are four main types for the nodes in a configuration tree: atoms, 
+// objects (which include arrays), DeferredConfigs (deferreds, for short), and 
+// resolvers.
+var nodeType = function(node) {
+  if (node instanceof Resolver) return 'resolver';
+  if (typeof node !== 'object' || !node) return 'atom';
+  if (node instanceof Date) return 'atom';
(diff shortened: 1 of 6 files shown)


Klortho:
I see it's failing to build on early versions of Node. I'll look into it when I get a chance.

Klortho:
Fixed.

markstos:
I've scanned this, and am in favor of merging. In favor:

   * Detailed docs both as code comments and a high level overview
   * Notable new test coverage
   * No regressions

The change /does/ affect people who don't use the feature, as the getters are added unconditionally. This could have a minor performance difference. As @Klortho said, this is probably isn't a difference that matters. 

If there is concern about the impact of this on users who don't use defer'ed values, we can add the extra check that I originally spec'ed, and only add the getters if there any deferred values to process.

Klortho:
Thanks for the review! You know, reading your comment, I think that it probably is a good idea to put the switch in so that it only performs the resolution if there's at least one deferred. I feel *pretty confident* that there won't be adverse side-effects, but it is a substantial change, and I don't know all the different ways people have cooked up to use this module, so, to be on the safe side, that flag is a good idea. Then, folks who never dreamed of using the deferred feature would assuredly not be impacted in any way.

Klortho:
I put it in: `deferredCount` -- exported in case it's convenient.

markstos:
Thanks. I have no further reservations about merging. It's up to @lorenwest to make the final call.

lorenwest:
Yes, an unfortunate name for what this does. For my sake and the benefit of others that may misunderstand - exactly what do these deferreds do? Is there a wiki page on their use?

markstos:
@lorenwest, because these only apply to the JavaScript file type, they are
described in the "JavaScript" section on the Configuration Files wiki page:

https://github.com/lorenwest/node-config/wiki/Configuration-Files#javascript-module---js

It's basically this paragraph, along with an example:

"In the JavaScript modules you have the option to define a configuration
value as function whose resolution will be deferred until the final merged
configuration structure is built. In the example above, a default subject
is provided that references another configuration value-- the site title.
Another configuration file may override the site title. Because the
resolution of email.subject is deferred, it would resolve to refer to the
overridden site title."

On Mon, May 9, 2016 at 2:43 PM, Loren West <notifications@github.com> wrote:

> Yes, an unfortunate name for what this does. For my sake and the benefit
> of others that may misunderstand - exactly what do these deferreds do? Is
> there a wiki page on their use?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/lorenwest/node-config/pull/318#issuecomment-217952137>
>



-- 
    Mark Stosberg
    Senior Systems Engineer
    RideAmigos

Klortho:
They merit more, I'd say. I think they can be super-useful, and was really excited to see that this modules has them. I'm in favor of renaming them, if you guys agree.

Klortho:
I have a write-up about the motivation for having them in my [settings-resolver repo readme](https://github.com/klortho/settings-resolver). Take a look -- maybe we could borrow some of that text for a wiki page here?

lorenwest:
Thanks for the link. I see the value, and while it's an unfortunate name I'm not sure we can change it without breaking installs.

I'm ok with the merge, and @markstos has reviewed the code so either he can merge, or I'll do a code review and merge it later.

Klortho:
As an aside, I first became interested in this approach recently when starting to use grunt. It defines [template strings](http://gruntjs.com/api/grunt.template) that do essentially what this does, but in a much more limited way. I got to thinking, why not use real functions; after all, it's JavaScript -- a very powerful *functional* language. I started to write a module that did this for grunt ([grunt-template-functions](https://github.com/Klortho/grunt-template-functions)) but it's not finished. Hmm ... now all of a sudden I'm thinking that maybe this could be factored into a separate module. But that's for another day, I suppose.

Cheers!

Klortho:
I think this needs one more test -- to make sure that it handles arrays the same way the existing version does. 

Also, I am a tiny bit concerned that it might break one edge case. If an existing user of deferreds creates complex JavaScript objects, with, for example, non-enumerable and/or inherited properties or methods, into their configs, and then use them *inside of other deferreds*, then they might not behave properly. The issue is that this uses Resolver objects inside of deferreds to represent (or proxy) the config objects. I think it's a real edge case -- any idea if it might be an issue? It's probably worth mentioning in the docs.

StyleT:
any updates here? May I help something?

markstos:
@StyleT Sorry, looks like this got stalled. Could you re-base the branch onto master and resolve the conflicts? Then we'll merge and release.

Klortho:
Sorry for the slow response. I'll take a look at this tonight.

markstos:
For this ending up here: I'm noting that @Klortho decided to go a different direction. Instead of refining his pull request for `node-config`, he created an alternate config library you can find at https://github.com/Klortho/config-one

iMoses:
@markstos I played with the code and found a way to solve this issue.
I haven't tested it thoroughly, but initial tests looks good.

It works like that:
1. Resolve all `DeferredConfig`
Resolving a `DeferredConfig` will return a promise instead of executing `func`. 
This promise has a custom method `exec`, which we will call after all deferred functions were resolved.
```
function deferConfig (func) {
  var obj = Object.create(DeferredConfig.prototype);
  obj.resolve = function(config, original) {
    var resolve, promise = new Promise(function(res) { resolve = res; });
    promise.exec = function() {
      resolve(func.call(config, config, original));
      return promise;
    };
    return promise;
  };
  return obj;
}
```
1. A. After each `DeferredConfig` is resolved we replace the resolve method with a new version which will `setPath` on the active configuration once the promise has been resolved.
```
if (prop[property] instanceof DeferredConfig ) {
  prop[property] = prop[property].resolve.call(completeConfig,completeConfig, prop[property]._original);
  prop[property].resolve = function() {
    prop[property].then(function (val) {
      util.setPath(t, path.concat(property), val);
    })
  };
}
```
2. Execute the `exec` method which will execute `func`, and resolve the promise with the results.
3. After all promises have been "executed" we execute the altered `resolve` method which will replace all promises with their final values.
This is async but if the user keeps all deferred functions "sync" (resolve within the next event loop) then most likely we won't experience a race condition (need to verify with tests).

Then you can do something like this:
```
var config = {
  a: defer(function (cfg) {
    return cfg.b.then(b => b);
  }),
  b: defer(function () {
    return 'value!';
  }),
};

// OR

var config = {
  value : 'value!',
  a: defer(async function (cfg) {
    const b = await cfg.b;
    return 'my '+b;
  }),
  b: defer(function (cfg) {
    return cfg.value;
  }),
};
```

markstos:
Closing due to lack of interest from anyone since 2017.

PR #322, opened 2016-05-18
Fix #321 - TypeError: Property description must be an object: undefined

### LINKED ISSUE #321: TypeError: Property description must be an object: undefined
With the latest version, 1.20.2, I get the following error when trying to use the util.extendDeep() function:

```
.../node_modules/config/lib/config.js:1420
        Object.defineProperty(mergeInto, prop, Object.getOwnPropertyDescriptor(Object(mergeFrom), prop));
               ^
TypeError: Property description must be an object: undefined
    at Function.defineProperty (native)
    at .../node_modules/config/lib/config.js:1420:16
    at Array.forEach (native)
    at util.extendDeep (.../node_modules/config/lib/config.js:1400:9)
```

This occurs when extending one config with another config on the 'get' function.
A simple fix would be to check if the Object.getOwnPropertyDescriptor() return value is not undefined before using it in the Object.defineProperty() call.

kgoerlitz:


kgoerlitz:
Added test case for the issue.

markstos:
I haven't studied the nuance here, but I'm merging this because the fix causes no regressions in the test suite, there's new test coverage, and it fixes a crashing regression that was just introduced. 

@lorenwest I recommend another release for this fix to a regression we accidentally introduced.

lorenwest:
I'll get it into NPM ASAP

lorenwest:
1.20.3 pushed to NPM. Thanks @kgoerlitz and @markstos for helping this through

PR #323, opened 2016-05-22
Fix prototype merge

tahoemph:
In 1.20.1 extendDeep would some things including methods on the prototype (e.g. has).  With 1.20.2 that when __lookupGetter__ was removed.  The fix in 1.20.3 didn't fix that although it did guard against the cases when getPropertyDescriptor returned undefined.  This change restores the current code structure where if what is being copied isn't one of the known types (including property descriptor) then it is simply copied.  This change also creates a test for this case.  This test passes in 1.20.1, not in 1.20.2/3, and then again after change is applied.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1416,11 +1416,10 @@ util.extendDeep = function(mergeInto) {
       }
 
       // Copy property descriptor otherwise, preserving accessors
-      else {
-        var descriptor = Object.getOwnPropertyDescriptor(Object(mergeFrom), prop);
-        if (descriptor) {
-          Object.defineProperty(mergeInto, prop, descriptor);
-        }
+      else if (Object.getOwnPropertyDescriptor(Object(mergeFrom), prop)){
+          Object.defineProperty(mergeInto, prop, Object.getOwnPropertyDescriptor(Object(mergeFrom), prop));
+      } else {
+          mergeInto[prop] = mergeFrom[prop];
       }
     }
   });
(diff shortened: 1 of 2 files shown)


lorenwest:
Is there any reason we're still supporting deepExtend vs. including it from [npmjs extend](https://www.npmjs.com/package/extend)? That package has 10 million downloads per month, and it's based on `jQuery.extend()`. I'd say it's been around the block.

tahoemph:
My goal here was to restore functionality that was removed in a patch version where I wouldn't take on any additional semantic skew changing the implementation to npmjs extend might entail.

markstos:
@lorenwest I thought I detected a bias against adding new dependencies, but I would be in favor of "outsourcing" a problematic bit of code with a dependency that provides a reliable dependency.  I would probably turn to Lodash for my own projects, but the `extend` module you pointed to looks a great, isolated dependency if it passes our tests.

https://lodash.com/docs#assignIn

markstos:
I just noticed there's a native `Object.assign()` method: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign Perhaps it would be helpful here.

markstos:
I just noticed there's a native `Object.assign()` method: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign Perhaps it would be helpful here.

lorenwest:
Thank you @tahoemph for discovering the issue, fixing the code, and including a test that validates the issue. I will get this in asap.

@markstos When node-config began in 2010, there was a bias against external module bloat. It was hard to tell the quality of an external module, and the style of the day was all-inclusive libraries like underscore. Today people are writing more independent modules, and it's easier to gauge the quality of a library by the history and number of downloads. Today I use lodash in all my apps, and they've made it easy to use parts of it without loading the entire library in memory - useful for libraries like node-config that need specific features.

lorenwest:
Published in 1.20.4

PR #326, opened 2016-05-31
Add XML config file support

rundef:


lorenwest:
This feature would require apps that use `.xml` configs to compile NodeJS with deasync - a module that's pure evil, and is breaking the travis build.

Maybe you could consider using the [xmldoc](https://github.com/nfarina/xmldoc) XML parser? It's not as widely used as xml2js, but the config package reads configs synchronously so they're available to your app without wrapping in callbacks or promises.

Whatever parser you use, you'll have to add documentation in the wiki to assure apps that make use of the `.xml` configs have it in their package.json file.

lorenwest:
XML is more expressive and differentiates data from metadata (elements vs. attributes), a wiki page explaining how the parser makes use of elements vs. attributes with examples would be welcome.

rundef:
Hi, thanks for the quick reply, I actually replaced xml2js by [xml2json](https://www.npmjs.com/package/xml2json) ... because it does return the data exactly like we want it (unlike xmldoc)

Let's see if the travis build works ... I will edit the wiki later on.

rundef:
Just edited the wiki [here](https://github.com/lorenwest/node-config/wiki/Configuration-Files) !

lorenwest:
Wiki looks great - thanks! No compiled node, travis tests pass, it's good for merging.  I'll get it in and published asap.

Thanks for the addition. It's kind of amazing it took so long for someone to need xml support enough to add it. Sign of the times...

lorenwest:
Just posted these changes on 1.21.0.  Thanks for making this addition!

Issue #309, opened 2016-04-16
SUPPRESS_NO_CONFIG_WARNING does not suppress the warning of NODE_APP_INSTANCE


adoyle-h:
![image](https://cloud.githubusercontent.com/assets/1998490/14581735/f4777142-0426-11e6-887b-b8e180fa5442.png)
## versions

node: v5.10.0
config: 1.20.1

markstos:
It is not trying to. See the docs for it: 

https://github.com/lorenwest/node-config/wiki/Environment-Variables#suppress_no_config_warning

Regarding Strict Mode, the message you posted links to a page with more about that.

adoyle-h:
> will be suppressed if no configurations were found when loading node-config. This is important for NPM modules using node-config to allow apps using these sub-modules the option of using node-config if they choose, or a different configuration system without getting annoying warning messages.

> specify the SUPPRESS_NO_CONFIG_WARNING environment variable so applications don't get No config files warnings if they aren't using node-config.

I think mismatched NODE_APP_INSTANCE has the same problem when use node-config for sub module configuration.
`SUPPRESS_NO_CONFIG_WARNING` is for keeping users from annoying warning messages thrown by node-config.

All `_warnOrThrow` invocations should be suppressed when `SUPPRESS_NO_CONFIG_WARNING` is set, shouldn't it?

markstos:
If you are getting warnings when there is no config file present and the SUPPRESS environment variable is set, then that seems like a bug. 

If you have /some/ config files and are getting a warning about the strictness check, that's intended.

adoyle-h:
Here is my situation:
If I write a library using node-config for configuration, and release it on npm.
Someone requires my library in his project while running it under the NODE_APP_INSTANCE set environment.
For example:

```js
// library.js, released on npm
process.env.SUPPRESS_NO_CONFIG_WARNING = 'y';
var config = require('config');
var myDefaultConfigs = {
    param1: "value1",
    param2: 22
};
var conf = config.util.extendDeep(myDefaultConfigs, {a: 1});
exports.config = conf;
```

```js
// project.js
process.env.NODE_APP_INSTANCE = 1;  // PM2 will set the NODE_APP_INSTANCE
require('library')
```

And someone run it under [pm2](https://github.com/Unitech/pm2): `pm2 start project.js`.

The warning messages may confuse the people.

markstos:
I agree this could be confusing. 

[...]

PR #343, opened 2016-08-22
Fix for #309: Interpret SUPPRESS_NO_CONFIG_WARNING more broadly

eloquence:
This implements the logic described in
https://github.com/lorenwest/node-config/issues/309#issuecomment-212011422

We also let the user know about whether strict mode is currently
on or off for relevant messages.

Additionally, runStrictnessChecks has been renamed to runSanityChecks,
since it is not only run in strict mode, and not all checks are
necessarily affected by strict mode.


lib/config.js
```diff
@@ -1653,17 +1673,25 @@ util.runStrictnessChecks = function (config) {
...
+      console.error(prefix+modeMessage);
+      if (beStrict) {
+        throw new Error(prefix+msg+' '+modeMessage);
+      }
+    }
+
+    // If we're not in strict mode or this is a message that's safe to disable, let the user
+    // know that they can turn off warnings altogether
+    if (!beStrict || neverThrow) {
+      console.error(prefix+'To disable this warning, set SUPPRESS_NO_CONFIG_WARNING in the environment.');
```
lorenwest: This is a generic function, yet you're outputting advise for only one of the invocations. 

This should be moved up to the place that calls _warnOrThrow for the specific invocation that would be affected by the SUPPRESS_NO_CONFIG_WARNING variable.

lorenwest:
I want @markstos to validate that it has met his criteria described in https://github.com/lorenwest/node-config/issues/309#issuecomment-212011422 before merging.

It also doesn't appear to have any tests validating the new functionality.

lib/config.js
eloquence: I don't understand what you mean by "you're outputting advise for only one of the invocations". The "To disable this warning, ..." message is shown for _all_ invocations if strict mode is off (since in those cases, it would have an effect), and independent of strict mode for the single case of "no config file", since it can be safely dismissed with strict mode on in that case (unchanged behavior).

eloquence:
Thanks @lorenwest! I've responded to your inline comment as well. If the code passes the smell test, I'll look into writing proper tests and may poke for help if I get stuck :)

lib/config.js
lorenwest: What I mean is, `_warnOrThrow()` is a generic function, but `SUPPRESS_NO_CONFIG_WARNING` is a specific message. That message makes no sense, for example, when calling `_warnOrThrow()` because NODE_ENV is ambiguous.

The "NO CONFIG WARNING" means to suppress the NO CONFIG warning if you have no configurations. It has nothing to do with the ambiguous NODE_ENV or any other warning that you may send to this generic function.

eloquence: Thank you @lorenwest, I think I see what you mean and misunderstood the original comment by @markstos (I missed the ", there are no config files," clause in the third bullet point). To clarify, we _do_ want the variable to suppress the specific warnings, even _if_ a default config is present, for the cases of NODE_ENV and NODE_APP_INSTANCE settings that don't match config files, correct? (That's what the original reporter wanted, and that's what I would also need for this to make sense for me.)

In that case, I'll refactor the code accordingly. Keep in mind that we still have three cases (no config at all, no instance config, NODE_ENV points to bad config) that require a user notice about the SUPPRESS environment variable.

We may want to consider, however, whether all of this really makes sense -- if, in future, we want to migrate to a single flag that suppresses all warnings during the initialization process, I would suggest rather preparing for that as a breaking change now (with a new environment variable name SUPPRESS_CONFIG_INIT_WARNINGS or such) since we're already inching up towards it anyway. That category of warnings does not seem so broad to me that this would be a bad idea.

If you agree with that as an alternative approach, I will open a separate ticket for it and you can help me make sure we handle the breaking change in a manner consistent with your release process. If you don't see the point of it, I'll just amend this pull request to handle the NO_CONFIG cases as described above.

markstos: > To clarify, we do want the variable to suppress the specific warnings, even if a default config is present, for the cases of NODE_ENV and NODE_APP_INSTANCE settings that don't match config files, correct?

Yes, if strict mode is not on. If strict mode is on, it's a fatal. error. This was points 1 and 2 of my "spec" about this should work:
1. If SUPPRESS_NO_CONFIG_WARNING is set and NODE_ENV does not match a config file, the warning will suppressed when strict mode is off, but still be a fatal error when strict mode is on.
2. If SUPPRESS_NO_CONFIG_WARNING is set and NODE_APP_INSTANCE does not match a config file, the warning will suppressed when strict mode is off, but still be a fatal error when strict mode is on.
3. If SUPPRESS_NO_CONFIG_WARNING is set, there are no config files, and NODE_ENV is set to "default" or "local", a warning will be suppressed when strict mode is off, but still be a fatal error when strict mode is on.

> we want to migrate to a single flag that suppresses all warnings during the initialization process

I don't want that. All of the warnings are there for a reason. I like the idea that you to explicitly suppress each class. With a single giant "disable all" flag, you might end up disabling a warning you've never seen before and didn't intend to disable-- some issue that would bite you only after you set the global "disable all warnings" flag. 

If anything, we should prioritize in a different direction: making it easy for your configuration to _quit generating warnings because the configuration is safe_.

eloquence: Okay, that makes sense -- will amend accordingly. It _may_ still make sense to change SUPPRESS_NO_CONFIG_WARNING to SUPPRESS_NO_CONFIG_WARNINGS at some point to make clear that it's a warning class, not a single warning.

markstos: Ah, agreed.

eloquence: Okay, do you just want me to do this name change as part of this PR? I don't believe formal deprecation code with a deprecation period is required, since users who previously used the old env name will just see a notice telling them about the new env name, and the warning is only shown once on load so there's no risk of huge log pollution. But we'd still update history, wiki, etc. Does that make sense?

eloquence: All done in the follow-up commit. Let me know if you prefer not to change the name for now and I'll take that bit out again. Will look into the tests and any other needed follow-up later today.

lorenwest:
I don't know how this PR got so far off track. We can't change the name of an existing environment variable, and I don't support an environment variable that suppresses all config warnings.

You may have misunderstood the reason for SUPPRESS_NO_CONFIG_WARNING. It's because we have this warning that's emitted if you have NO CONFIGS defined. That environment variable is specifically to make it so that ONE warning isn't emitted.

Unless I'm mistaken, the change at https://github.com/lorenwest/node-config/pull/343/files#diff-f7bf2b665273dfa66ffa4ad7c0a52bb9R1647 would send a warning if your NODE_ENV is `default` or `local` - why would you do this? I believe it's lack of understanding of what @markstos was asking, and has me questioning your understanding of the code well enough to make such a big change.

Let's get back to the basics. The thread in #309 discussed allowing NO CONFIG warnings under certain situations to be suppressed because it was meaningless for pm2 users. One comment in the thread described using node-config in sub-modules improperly (we have a different mechanism for [sub module configuration](https://github.com/lorenwest/node-config/wiki/Sub-Module-Configuration)).

I appreciate the effort you've put in to this, and want to understand your use case better. Are you using pm2 with sub-module configuration?

lorenwest:
I apologize for that last post. It was my lack of understanding, not yours. It took me a while to fully understand the changes you had made.  My only comment is we can't change the name of the environment variable because others are already using the singular version.

Man, I should stop doing late night code reviews.

eloquence:
No problem, thanks for the late night review. :) I figured it was a relatively safe name change if it's properly announced, but I can change it back. Will add some tests now, which I should have done to begin with and which would have helped with the review and conversation - my bad. :)

eloquence:
Done, let me know if these tests make sense to you. I took the liberty of removing an unnecessary  assignment from the previous batch for consistency.

markstos (approved):
Sorry for the... half decade... delay here.

Looks good. If the original author anyone else wants to resolve the conflicts here and squash it into a single commit, I think it can be merged.

markstos:
This could be expanded to also fix #644-- warn or throw in the case when the host is unfortunately named "default".

eloquence:
Thanks! Oh boy, it's been a while, will need to stare at this for a little before I can meaningfully test & rebase it - will try to poke at this over the weekend :)

JonaC22:
Any news about this? It would be much appreciated. Thank you!

markstos:
@JonaC22 Still waiting for a volunteer to rebase and resolve the conflicts here and squash it into a single commit. It could be you.

markstos:
Closing. Superceded by #680

Issue #276, opened 2015-12-14
Add env var to search additional config directories in order to override values found in $NODE_CONFIG_DIR


matthewadams:
I've authored my app so that the default configuration values in `config/default.json` make the app runnable (on a dev's machine, for instance).  However, in other environments (unit-test, qa, staging, production), I want to override config values, but not have to copy my environment-specific config file into my app's config dir but still use the defaults found in my app's `config/default.json`.

According to the docs, setting `NODE_CONFIG_DIR` changes the config directory that's searched for config files -- I'm assuming _all_ config files.  I want to specify an extra directory, _in addition_ to `config` so that I don't have to copy, say, `production.json` into `config`.  I want to
1. leave `production.json` in its well-known place on the production server, say, `/foo/bar/production.json`,
2. have all defaults in my regular config directory, say `/sna/fu/app/config/default.json`,
3. set `NODE_ENV` to `production`,
4. set `NODE_CONFIG_DIRS` to `/foo/bar`,
5. fire up my app and have it pick up `/foo/bar/production.json` _and then_ `/sna/fu/app/default.json`.

By using `NODE_CONFIG_DIRS` (with an `S`), `config`
1. leaves legacy behavior of `NODE_CONFIG_DIR` but allows additional directories to be searched _before_ `$DEFAULT_NODE_CONFIG` is searched, and
2. allows me to copy my application into a directory and just run, no additional copying required (if `NODE_CONFIG_DIRS` is already set).  Yay, one less production deployment step!

My requirement is only for one additional directory, but I don't see why multiple additional directories couldn't be supported using either JavaScript array notation (`like ['/foo/bar','/goo/far']`) or a single string with platform-specific path separators (like `/foo/bar:/goo/far`on _n_x or `\foo\bar;\goo\far` on Win) where first config file found wins.

markstos:
Why not just cp production.json into config/ as part of deployment?

Also, why not check production.json into git?

I gave up on having configs that only exist in production when I
accidentally deleted production and the backup of production at the same
time.

lorenwest:
Node-config is designed to keep as many configurations in source code control as possible.  It allows for deployment specific configurations, even machine specific configurations.

For the few configs that truly are dynamic, those are generally passed in as environment variables.

I'm not a fan of copying configs from the filesystem into the app /config directory, because everything under the app should be a representative of what's found in your source control and in npm publish.

matthewadams:
[...]

PR #350, opened 2016-10-01
util.loadFileConfigs: support optional source dir

wmertens:
Example use case: load module defaults which you then merge with util.setModuleDefaults
```jsx
import config from 'config'
const ourConfigDir = path.join(__dirname, 'config')
const baseConfig = config.util.loadFileConfigs(ourConfigDir)
config.util.setModuleDefaults('MyModule', baseConfig)
```

See #276

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -625,7 +625,7 @@ util.getConfigSources = function() {
  * @method loadFileConfigs
  * @return config {Object} The configuration object
  */
-util.loadFileConfigs = function() {
+util.loadFileConfigs = function(configDir) {
 
   // Initialize
   var t = this,
@@ -633,7 +633,7 @@ util.loadFileConfigs = function() {
 
   // Initialize parameters from command line, environment, or default
   NODE_ENV = util.initParam('NODE_ENV', 'development');
-  CONFIG_DIR = util.initParam('NODE_CONFIG_DIR', Path.join( process.cwd(), 'config') );
+  CONFIG_DIR = configDir || util.initParam('NODE_CONFIG_DIR', Path.join( process.cwd(), 'config') );
   if (CONFIG_DIR.indexOf('.') === 0) {
     CONFIG_DIR = Path.join(process.cwd() , CONFIG_DIR);
   }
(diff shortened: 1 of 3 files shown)


lib/config.js
```diff
@@ -701,42 +701,45 @@ util.loadFileConfigs = function() {
...
   }
 
   // Override with environment variables if there is a custom-environment-variables.EXT mapping file
   var customEnvVars = util.getCustomEnvVars(CONFIG_DIR, extNames);
   util.extendDeep(config, customEnvVars);
 
-  // Place the mixed NODE_CONFIG into the environment
-  env['NODE_CONFIG'] = JSON.stringify(util.extendDeep(envConfig, cmdLineConfig, {}));
-
   // Extend the original config with the contents of runtime.json (backwards compatibility)
```
markstos, line 743: Isn't there a logic change here?

It appears that in the original logic, 'NODE_CONFIG`was set in the environment *after* the config was extended with`customEnvVars`, but after this change it appears that`NODE_CONFIG`is set in the environment *before* he config was extended with`customEnvVars`.

This may not be a case that we had a regression test for.

wmertens: I moved that up so that it could live in the same `if` block as the rest of the `NODE_CONFIG` support. I did not consider the case that the customEnvVars might be referring to NODE_CONFIG. Is that even possible, given that it is a JSON string? Did not look into that deeply yet.

markstos: On second look, this order change is not a problem. The setting of `NODE_CONFIG` in the environment that was moved does not reference the `customEnvVars` variable, nor the `config` variable, which are the ones handled between the old location of the code and then new location of the code. 

In this case, I agree with @lorenwest that this is ready to merge. Then the method should be documented on the wiki. https://github.com/lorenwest/node-config/wiki/Using-Config-Utilities

wmertens:
Documented at https://github.com/lorenwest/node-config/wiki/Using-Config-Utilities#loadfileconfigsdirectory

lorenwest:
Published 10/25 in 1.22.0

PR #351, opened 2016-10-21
original/previous value for deferredConfig

simon-scherzinger:
I would like to suggest this PR to extend functionality for deferredConfigs. This PR is fully backward compatible.

When merging a config file which has deferred values the value from the previous file is saved into the DeferredConfig object. Later when deferred values are resolved this saved value is passed to the resolver function.

I am currently developing a deferred resolver which allows me to load values from etcd – hopefully I will be able to release this in the next week. Yes, it is synchronously but this is ok in my context. When no value is in etcd I wanted to get the last one so I did this little change.

Maybe some other people are in need of this? :)

Here is an example:

```yaml
# default.yml
someValue: default
```

```js
// local.js
module.exports = {
  someValue: defer(function (cfg, original) {
    return getSomeFancyConfigOrUndefined() || original;
  }
}
```

When `getSomeFancyConfigORUndefined` is not able to get a value, we are able to use the `original` value now.

### DIFF
--- a/defer.js
+++ b/defer.js
@@ -1,7 +1,7 @@
 // Create a deferredConfig prototype so that we can check for it when reviewing the configs later.
 function DeferredConfig () {
 }
-DeferredConfig.prototype.resolve = function (config) {};
+DeferredConfig.prototype.resolve = function (config, original) {};
 
 // Accept a function that we'll use to resolve this value later and return a 'deferred' configuration value to resolve it later.
 function deferConfig (func) {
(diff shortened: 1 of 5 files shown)


lorenwest:
@markstos, you're the one that understands this deferred mechanism best. I'm comfortable with it if you are.

markstos:
Merged. @simon-scherzinger could you update the appropriate wiki page to document this?

We'll try to get a release out soon to keep the released version and wiki in sync.

simon-scherzinger:
Thanks for accepting.

Updated the wiki page just [now](https://github.com/lorenwest/node-config/wiki/Configuration-Files/57973a5e8ab3565a109bfd16111451257386114c).

lorenwest:
Published 10/25 in 1.22.0

Issue #118, opened 2014-05-21
Support CSON files


djensen47:
CoffeeScript support is great but what would be really nice is CSON file support. See this project: https://github.com/bevry/cson

PR #358, opened 2016-11-11
Play nice with transpilers by using variable requires (fixes #345)

### LINKED ISSUE #345: devDependencies errors on ES6
I've used **config** before on a previous project and things worked great. I really like the library for it's simplicity and ease of use. It's like the **underscore** of configurations.
Anyway, I recently decided to create a new project based on the old one, but this time using ES6 syntax, that would then get transpiled to ES5 using **gulp** and **babel**. 
Sadly, things didn't work out as I expected. Attempting to transpile the code or even to run it (using `npm run start`) throws an error that several modules were not found. After some inspection, I found out that these are ALL modules listed as **_devDependencies_** for **config**.
The **_devDependencies_** aren't installed using `npm i` (but I didn't expect them to)...but they also weren't installed in the old project (yet that still worked fine).
Any ideas on how to resolve this?

blahah:
As discussed in #345 this PR moves all conditional requires in `lib/config.js` to use a variable module name. This allows transpilers like Webpack to properly resolve the dependency graph.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -537,7 +537,7 @@ util.makeImmutable = function(object, property, value) {
   for (var i = 0; i < properties.length; i++) {
     var propertyName = properties[i],
         value = object[propertyName];
-    
+
     if (!(value instanceof RawConfig)) {
       Object.defineProperty(object, propertyName, {
         value: value,
@@ -872,7 +872,8 @@ util.parseFile = function(fullFilename) {
         //  Coffee = require("coffee-script");
         //}
 
-        Coffee = require("coffee-script");
+        var dep = "coffee-script"
+        Coffee = require(dep);
 
         // coffee-script >= 1.7.0 requires explicit registration for require() to work
         if (Coffee.register) {
@@ -883,7 +884,8 @@ util.parseFile = function(fullFilename) {
       configObject = require(fullFilename);
     }
     else if (extension === 'iced') {
-      Iced = require("iced-coffee-script");
+      var dep = "iced-coffee-script"
+      Iced = require(dep);
 
       // coffee-script >= 1.7.0 requires explicit registration for require() to work
       if (Iced.register) {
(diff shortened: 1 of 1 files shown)


lib/config.js
```diff
@@ -883,7 +884,8 @@ util.parseFile = function(fullFilename) {
       configObject = require(fullFilename);
     }
     else if (extension === 'iced') {
-      Iced = require("iced-coffee-script");
+      var dep = "iced-coffee-script"
```
lorenwest: Javascript hoists variables, and having individual `var dep` gives the misleading impression that these are different things, when they are all the same variable.

A better pattern would be to define `var dep = undefined;` at the top of the function and use that variable without (misleadingly) re-defining it in each instance.

Even better would be to have a block at the top of the file like this

```
// Define soft dependencies so transpilers don't include everything
var ICED_COFFEE_SCRIPT = "iced-coffee-script";
var JS_YAML = "js-yaml"
...
```

Then simply use these variables instead of the hardcoded strings `require(ICED_COFFEE_SCRIPT);`

blahah: +1 to that. I blindly implemented the isolated solution given for one variable in the issue thread. I'll push an update now.

jonohodgson:
I've taken this config.js but still getting
ERROR in ./~/config/lib/config.js
Module not found: Error: Cannot resolve module 'fs'

Is this working for others?

blahah:
@jonohodgson I eventually discovered that it's extremely hard to get node-config working with Webpack (or any transpiler) in the browser environment in any meaningful way. This is because there is no `fs` module in the browser, and it's impossible to fully implement one.

In Webpack you can avoid that specific error using `node: { fs: "empty" }` (see [the Webpack docs](https://webpack.github.io/docs/configuration.html#node)) which will give a no-op on any fs call. However, you will still face the problem that your config file will not be found by `node-config`, because it can't access the real filesystem. You can hack your config files into an [in-memory filesystem polyfill](https://github.com/webpack/memory-fs), but at this point you may as well just have a config object in memory.

Ultimately we settled on exposing a shimmed CONFIG global using Webpack [EnvironmentPlugin](https://webpack.github.io/docs/list-of-plugins.html#environmentplugin). An alternative that we considered but ultimately didn't need was [node-config-loader](https://github.com/zerkalica/node-config-loader).

botagar:
Hi. I was wondering what the progress of this PR is, and If I should create a similar one if this branch is not being worked on anymore.
This fix is something I need, and I would like the freedom to be able to clear my packages and re-install at will.

Cheers~!

jdmarshall:
This PR appears to be answered by #118

markstos:
@jdmarshall #118 is an old issue about CSON support. Try again?

markstos:
Given the comment about that says essentially "this wasn't going to work in the end anyway": https://github.com/lorenwest/node-config/pull/358#issuecomment-261520186 I'm closing this.

Issue #368, opened 2016-12-21
Wish: support Electron environment


dw1284:
This issue seems to be related to the fact that **process.cwd()** doesn't return what you think it does when running via Electron. Replacing **process.cwd()** with **require('Electron').app.getAppPath()** fixes the issue, but that only works if you have Electron included as a dependency. Maybe add an if check to determine if Electron is present, and if so, use the getAppPath() function, otherwise use the regular process.cwd() function.

markstos:
We would be open to running in the Electron environment. First, we need a way to explicitly detect that we are in the electron environment. How does one detect they are running in the electron environment? Can you provide examples of modules that handle this case in both regular Node.js and the Electron environment?

This enhancement request is very closely related to #367.  The same approach should be used in both cases. 

The `node-config` maintainers are using `node-config` inside Electron but we consider reviewing a PR if someone gets this working.

dw1284:
**Update:** Loren indicated, in this thread https://github.com/lorenwest/node-config/pull/369, that this fix isn't going to be implemented.

Loren did supply a legit work-around. You can find it in https://github.com/lorenwest/node-config/pull/369, however, I threw together a new module which wraps node-config and implements the work-around for you. This new module supports electron out of the box:

https://www.npmjs.com/package/electron-node-config

You can use the new module, or implement Loren's work-around yourself. Your choice.

I'm closing this since we now have multiple ways to deal with the issue, and there are no plans to change node-config as far as I can tell.

Thanks everybody.

PR #369, opened 2016-12-22
Support Electron environment

dw1284:
https://github.com/lorenwest/node-config/issues/368

Detect when running in an Electron environment. If running in Electron, retrieve the working directory from Electron instead of from the main process.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -638,11 +638,20 @@ util.loadFileConfigs = function(configDir) {
   var t = this,
       config = {};
 
+  // Detect when running in the Electron environment
+  // https://github.com/lorenwest/node-config/issues/368
+  var workingDir;
+  if (process && process.versions && process.versions.electron) {
+    workingDir = require('electron').app.getAppPath();
+  } else {
+    workingDir = process.cwd();
+  }
+
   // Initialize parameters from command line, environment, or default
   NODE_ENV = util.initParam('NODE_ENV', 'development');
-  CONFIG_DIR = configDir || util.initParam('NODE_CONFIG_DIR', Path.join( process.cwd(), 'config') );
+  CONFIG_DIR = configDir || util.initParam('NODE_CONFIG_DIR', Path.join( workingDir, 'config') );
   if (CONFIG_DIR.indexOf('.') === 0) {
-    CONFIG_DIR = Path.join(process.cwd() , CONFIG_DIR);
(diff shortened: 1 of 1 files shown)


markstos:
So to confirm: Has this patch been tested to work in an Electron environment?

`node-config` also uses the `fs` and `os` modules. Do this work normally under Electron?

dw1284:
I can confirm that this is working in an Electron environment. I also tested to ensure that it is still working in a non-Electron environment. As far as I can tell, the fs and os modules are working just fine. I'm assuming the whole thing would stop working if those weren't available.

markstos:
Thanks.  I'm in favor of including a version of this. When I have time, I'll revise to abstract the `cwd` calculation into it's own subroutine.

dw1284:
Sounds great! Thanks for taking the time to review this one.

joshuacurtiss:
Any chance at getting this merged and closing #368? Many thanks for a great library.

lorenwest:
Sorry this dropped off our radar. 

I'm not a fan of adding application specific code to a generalized library. This should work for Electron users:

```
if (process.versions && process.versions.electron) {
  process.env.NODE_CONFIG_DIR = require('electron').app.getAppPath() + '/config';
}
```

This code fragment would be placed above the first call to `require('config')` in your application startup module. Subsequent calls in other modules will get the correct config.

I don't have Electron, so I can't confirm. @dw1284 or @joshuacurtiss could you confirm that this works for your application?

Thanks for bumping this topic,
-Loren

dw1284:
Understood. I would suggest, at the very least, adding your snippet to the documentation.

This wasn't the easiest issue to deal with because everything works fine via npm run and VSCode debugging, however, as soon as you package into an Electron executable, the whole app breaks. I spent a few hours trying track down the nature of the problem. You could prevent other devs from spending those hours by adding the snippet to the docs.

Thanks for taking a look at this issue and providing a solution. Much appreciated.

dw1284:
I can confirm that the code snippet works. I have tested in an Electron app.

Also, for anybody who doesn't want to mess with adding the snippet to their code, I threw together another module called electron-node-config which adds the snippet for you. It is wraps node-config and supports Electron out of the box.

https://www.npmjs.com/package/electron-node-config

lorenwest:
Thank you!

joshuacurtiss:
Thank you all, you are heroes and saints!  😉

PR #375, opened 2017-01-25
Add support for configuration files stored with git-crypt.

cunneen:
currently node-config crashes with a parse error if, for example, you have a production.json file encrypted with [git-crypt](https://github.com/AGWA/git-crypt) due to the file containing sensitive information. This is because such files are binary files.

This PR adds transparent support for git-crypt as follows:

* If none of your configuration files are encrypted with git-crypt, node-config behaviour is unchanged.
* If you *DO* have a configuration file encrypted with git-crypt:
    * If the ```CONFIG_SKIP_GITCRYPT``` environment variable is truthy, the file is ignored and a warning is emitted to the console.
    * If the ```CONFIG_SKIP_GITCRYPT``` environment variable is not set or is falsy, an additional error message is emitted to the console just prior to the (existing) parse error.

Of course, if you are using git-crypt but your git-crypt files are unlocked, node-config works perfectly already. This issue only applies to developers who do not have the files unlocked.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -29,7 +29,8 @@ var DEFAULT_CLONE_DEPTH = 20,
     privateUtil = {},
     deprecationWarnings = {},
     configSources = [],          // Configuration sources - array of {name, original, parsed}
-    checkMutability = true;      // Check for mutability/immutability on first get
+    checkMutability = true,      // Check for mutability/immutability on first get
+    gitCryptTestRegex = /^.GITCRYPT/; // regular expression to test for gitcrypt files.
 
 /**
  * <p>Application Configurations</p>
@@ -647,6 +648,7 @@ util.loadFileConfigs = function(configDir) {
   APP_INSTANCE = util.initParam('NODE_APP_INSTANCE');
   HOST = util.initParam('HOST');
   HOSTNAME = util.initParam('HOSTNAME');
+  CONFIG_SKIP_GITCRYPT = util.initParam('CONFIG_SKIP_GITCRYPT');
 
   // This is for backward compatibility
   RUNTIME_JSON_FILENAME = util.initParam('NODE_CONFIG_RUNTIME_JSON', Path.join(CONFIG_DIR , 'runtime.json') );
@@ -827,13 +829,14 @@ util.parseFile = function(fullFilename) {
   var t = this,
       extension = fullFilename.substr(fullFilename.lastIndexOf('.') + 1),
       configObject = null,
-      fileContent = null;
+      fileContent = null,
+      stat = null;
 
   // Return null if the file doesn't exist.
   // Note that all methods here are the Sync versions.  This is appropriate during
   // module loading (which is a synchronous operation), but not thereafter.
   try {
-    var stat = FileSystem.statSync(fullFilename);
+    stat = FileSystem.statSync(fullFilename);
     if (!stat || stat.size < 1) {
       return null;
     }
@@ -852,6 +855,15 @@ util.parseFile = function(fullFilename) {
 
   // Parse the file based on extension
   try {
+
+    // skip if it's a gitcrypt file and CONFIG_SKIP_GITCRYPT is true
+    if (CONFIG_SKIP_GITCRYPT) {
+      if (gitCryptTestRegex.test(fileContent)) {
+        console.error('WARNING: ' + fullFilename + ' is a git-crypt file and CONFIG_SKIP_GITCRYPT is set. skipping.');
+        return null;
+      }
+    }
+
     if (extension === 'js') {
       // Use the built-in parser for .js files
       configObject = require(fullFilename);
@@ -895,6 +907,9 @@ util.parseFile = function(fullFilename) {
     }
   }
   catch (e3) {
+    if (gitCryptTestRegex.test(fileContent)) {
+      console.error('ERROR: ' + fullFilename + ' is a git-crypt file and CONFIG_SKIP_GITCRYPT is not set.');
+    }
     throw new Error("Cannot parse config file: '" + fullFilename + "': " + e3);
   }
 


markstos:
@cunneen Help me understand what was causing the crash. Was the issue that  git-crypt files are stored in the ./config directory and have names like `.GITCRYPT.develop.json`. Our file loader then attempts to load the file and parse it as JSON because it has a JSON extension, so then it crashes?

markstos:
I've in favor of supporting ways to manage sensitive data with node-config. I expect there are multiple solutions that people will want to use for this, just like there are many config file formats people want to use. Solutions that could serve more than secure storage solution are preferred, but having a specific solution to consider is a great starting point for discussion.

lorenwest:
Help me understand what this does. It appears to simply overlook certain files. Why do you need code to do this? Can't these files be stored some place else?

lorenwest:
Also, these aren't JSON files, but they have a JSON extension. That's a smell.

lorenwest:
As @markstos mentions, I would support a discussion of different mechanisms for securing configuration data. If changes were necessary as part of one of these mechanisms, I would support those changes.

These changes appear to solve a particular problem with one of the mechanisms, but without any context for how it helps people secure their configuration data, it's premature to consider it for a pull request. Closing the pull request for now, and if a discussion continues on how we can help people secure config data, I'd be open to future pull requests.

cunneen:
Hi @lorenwest and @markstos  : there's no 'smell'; they're encrypted JSON files, using git-crypt (which I linked in the pull request comments). git-crypt has great documentation so I didn't see the need to paraphrase it, but obviously there is some confusion so here is one scenario where it is very useful: 

* Loren, Mark and Mike are all working on a project, using git. There are server deployments involved. Mike is a contributor, and doesn't need or want to see production passwords, API secret keys etc. Loren and Mark both need access to the sensitive info.
* Loren uses git-crypt so that the production.json file is encrypted for Mike, but is plain JSON for himself and Mark:

    ```
    ```

cunneen:
Hi @lorenwest and @markstos  : there's no 'smell'; they're encrypted JSON files, using git-crypt (which I linked in the pull request comments). git-crypt has great documentation so I didn't see the need to paraphrase it, but obviously there is some confusion. So here is one scenario where it is very useful: 

* Loren, Mark and Mike are all working on a project, using git. There are server deployments involved. 
* Mike is a contributor, and doesn't need or want to see production passwords, API secret keys etc. 
* Loren and Mark both maintain the servers and need access to the sensitive info.
* Loren uses git-crypt so that the ```production.json``` file is encrypted for Mike, but the same file is plain JSON for himself and Mark:

    ```sh
    # initialise the repository to use git-crypt
    git-crypt init 
    
    # Loren adds his own public GPG key (already on 
    #  his GPG keychain) as a trusted user
    git-crypt add-gpg-user lorenwest@example.com 
    
    # Loren imports Mark's public GPG key to his own 
    #  GPG keychain (Mark probably emailed this to 
    #  Loren)
    gpg --import markstos_pub.gpg 
    
    # Loren adds Mark's public GPG key as a trusted
    #  git-crypt user
    git-crypt add-gpg-user markstos@example.com
    
    # Loren adds the config/production.json file
    #  to the .gitattributes file, so git-crypt will
    #  manage the encryption / decryption where 
    #  necessary.
    # This adds the following line to the 
    #   .gitattributes file:
    # config/production.json filter=git-crypt diff=git-crypt
    echo 'config/production.json filter=git-crypt diff=git-crypt' >> .gitattributes 

    # Now Loren adds a new config/production.json file 
    #  to his repo
    echo '{foo:"secret"}' > config/production.json
    git add config/production.json
    git commit -m "production settings"
    git push origin
    
    ```
    
* On the git server, ```config/production.json``` is now encrypted.
* Now Mark clones or pulls the repo. Initially, ```config/production.json``` is encrypted for him too. But his public key is listed as trusted, and he has the corresponding private key on his keyring. So he can unlock all encrypted files with a single command:
    
    ```sh
    git-crypt unlock
    ``` 
    
    Mark is prompted for his private key password.
* From now on, Mark and Loren's workflow is unchanged from normal. On their machines, ```config/production.json``` is just a normal JSON file.
* Now Mike wants to make a contribution. But he has a problem: when he runs the node program, node-config tries to read ```config/production.json```, which on his machine is not a valid JSON file since it's encrypted. The app crashes.
* With this pull request, Mike can specify that any git-crypt encrypted files are skipped, via an environment variable, as follows:

	```sh
	NODE_ENV=production CONFIG_SKIP_GITCRYPT=1 node foo.js
	```

lorenwest:
When I specify `NODE_ENV=production` I've set an expectation that the environment is production (or whatever environment I've specified). If that file is encrypted, I'm unable to run in the environment I've just asked to run under. Continuing in this situation is not appropriate, because it's unable to be configured for the requested environment.

If these encrypted files make their way into the production environment, servers would be started with incorrect configurations. The current behavior is correct - fail fast when specifying an environment that you don't have permission to run within.

The suggested workflow seems reasonable for encrypting secrets using git-crypt. The only issue I have is in the passive behavior when attempting to run an encrypted configuration.

cunneen:
Well, yes. If you want to encrypt your production files (like I do), of course you'll need to decrypt them at some point before running them. So you'd add a 'deployment' GPG public key on your production server.

I agree the current behaviour is correct, which is why my pull request doesn't change the current behaviour without an explicit override. 

There are cases where you want to mimic a production environment's behaviour without needing the sensitive details. The PR allows this.

This change is entirely 'opt-in'. If you don't want to use git-crypt, you won't be affected at all.

markstos:
>  there's no 'smell'; they're encrypted JSON files.

Developers and tools have a reasonable expectation that a file that ends in ".json" is a JSON file. `git-crypt` apparently breaks the file-extension convention by adding a *prefix* to note that the file is encrypted. If `git-crypt` had just followed file-extension convention and named the encrypted file `production.json.gpg`, `node-config` would automatically skip over the file and no change would needed.

So yes, `git-crypt`'s use of file-prefixing rather than file-suffixing is a smell.

cunneen:
I guess that's a valid opinion. I'm not really here to argue the merits of git-crypt. I need support for it. 

I'd much rather have it in node-config (and can't really understand the objections to a code change that has no negative impact whatsoever). 

For the benefit of anyone who comes across this PR wanting git-crypt support, I've published it to npm as [config-gitcrypt](https://www.npmjs.com/package/config-gitcrypt).

markstos:
@lorenwest One thing I see in the PR is that the files being worked-around start with a leading dot, indicating a private file: ".GITCRYPT...". What about just updating the config file reading code to skip *all* files that start with a leading dot? It would be a backwards-incompatible change, but I can't think of any reason why someone would be naming their config files with a leading dot, but here's a case where the leading dot as properly used to used to indicate a hidden or private file.

This seems to be all the PR @cunneen does is to skip these files, it's just specific to this particular kind of "dot" file instead of all dot files.

cunneen:
It's not filenames that start with a dot. It's the file contents, beginning with a specific non-ASCII character that I'm ignoring in the pattern, then GITCRYPT. The file contents are already in a buffer.

markstos:
I see now that git-crypt is doing in-place file encryption. The fact that encryption is in-place is somewhat unfortunate for playing nice with other tools which have reasonable expectations about file extensions mean.

An approach that could work with node-config is to store the maybe-encrypted files with a different extension that we don't try automatically parse.  Name it `production.json.git-crypt` or `production.json.gpg`

Then use a build tool to copy an unencrypted file to production.json when it's suitable to do so, but node-config runs.

Or, you could create `config/production.js`. Since the file is JavaScript, you can add the logic you need to pull in `production.json.git-crypt` contents when appropriate.

Depending on the approach, the result may be something that can be published and shared as a kind of "node-config pre-processor".

This kind of a "plug-in" approach would scale to any number of other encryption systems, without `node-config` requiring in-core patches for each different approach.

cunneen:
Are we discussing a product roadmap, or this PR ? This suits my needs, and I thought others might find it useful. That's all.

lorenwest:
If it's to fulfill one person's needs, forking is the right mechanism. We have to consider the overall impact to everyone, both positive and negative. Sure, there may be no negative impact (although every code change has maintenance overhead), but if there's no positive impact or it goes against some other convention, it's a no-go.

I'd rather see a procedural solution than a code solution. Have you considered the suggestions that @markstos made regarding file naming?

cunneen:
The reason we went with git-crypt was its transparent in-place encryption i.e. deployment tools and developer workflows "just work" without having to write additional scripts etc.

The reason we went with node-config is it allows us to split out the production config, and protect the sensitive info within via git-crypt.

@markstos 's suggestion may work for some, but it's too much manual overhead for us. The usability and simplicity of this system is hard to beat. Developer tools, Build tools, CI tools and deployment tools don't need extra configuration.

cunneen:
Certainly. It'll have to wait a few days, but I'm more than happy to do this.

PR #376, opened 2017-01-30
Git crypt support

cunneen:
Reissue PR #375 with tests. Adds git-crypt support. See #375 for discussion, and [the wiki entry](https://github.com/lorenwest/node-config/wiki/Securing-Production-Config-Files) for more info.

PR #377, opened 2017-02-01
fix bug when user does export CONFIG_SKIP_GITCRYPT env

ramimoshe:
Fix exception:

ReferenceError: CONFIG_SKIP_GITCRYPT is not defined
at util.loadFileConfigs (/usr/local/jive/servers/realtime/lib/node_modules/rt/lib/v2/services/connector/node_modules/config/lib/config.js:651:24)
at new Config (/usr/local/jive/servers/realtime/lib/node_modules/rt/lib/v2/services/connector/node_modules/config/lib/config.js:122:27)
at Object.<anonymous> (/usr/local/jive/servers/realtime/lib/node_modules/rt/lib/v2/services/connector/node_modules/config/lib/config.js:1735:31)

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -648,7 +648,7 @@ util.loadFileConfigs = function(configDir) {
   APP_INSTANCE = util.initParam('NODE_APP_INSTANCE');
   HOST = util.initParam('HOST');
   HOSTNAME = util.initParam('HOSTNAME');
-  CONFIG_SKIP_GITCRYPT = util.initParam('CONFIG_SKIP_GITCRYPT');
+  CONFIG_SKIP_GITCRYPT = util.initParam('CONFIG_SKIP_GITCRYPT', 1);
 
   // This is for backward compatibility
   RUNTIME_JSON_FILENAME = util.initParam('NODE_CONFIG_RUNTIME_JSON', Path.join(CONFIG_DIR , 'runtime.json') );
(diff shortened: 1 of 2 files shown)


markstos:
@ramimoshe your proposed fix causes a test suite failure. Could you investigate?

ramimoshe:
Ok, I can add one more test case with unset env
But the problem that I have is:
I updated my project to the latest node-config 
so I dont have this env 
and then I tried to run my project, and I got an exception

you can try it in your project

markstos:
@ramimoshe I believe you there's a significant issue with the latest release because the test coverage wasn't sufficiently good last time around. 

That's why I'm especially interested that the follow-up fix is of high quality with good test coverage and commit messages for the next update to be especially clear what's happening.

lorenwest:
It appears this only happens in runtime environments that restrict the use of global variables, which is why Travis didn't catch it. The problem was a leaked variable into global scope. I think I've fixed the problem, but @ramimoshe would you mind verifying that 1.25.1 resolves your issue?

cunneen:
This looks like a good test to add (CONFIG_SKIP_GITCRYPT=0). I agree with @markstos that it should be in addition to the existing tests, rather than replacing them.

Apologies for the inconvenience caused by omitting the var declaration in PR #376 ; for some reason I had the declaration uncommitted in my local repo. Thanks @lorenwest for cleaning it up.

Issue #329, opened 2016-06-09
"Cannot redefine property" error using util.extendDeep to extend a config


rpaterson:
Using the `Config.util.extendDeep(...)` utility errors when you're extending using config objects and there's an override.  This issue was introduced in v1.21.0 and does not error in previous versions.

default.json:

``` json
{
    "conf": {
        "key": "value"
    },

    "overrideConf": {
        "key": "override value"
    }
}
```

Code:

``` Javascript
const CONFIG = require('config');
console.log(
    JSON.stringify(
        CONFIG.util.extendDeep(
            {},
            CONFIG.get('conf'),
            CONFIG.get('overrideConf')
        )
   )
);
```

In earlier versions that would have output `{"Key":"override value"}` but in v1.21.0 it throws an error:

```
TypeError: Cannot redefine property: key
    at Function.defineProperty (native)
    at /home/rpaterson/programming/test/node_modules/config/lib/config.js:1434:18
    at Array.forEach (native)
    at util.extendDeep (/home/rpaterson/programming/test/node_modules/config/lib/config.js:1414:9)
    at repl:1:52
    at REPLServer.defaultEval (repl.js:262:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:431:12)
    at emitOne (events.js:82:20)
```

matthewadams:
I'm having the same issue.  If there is no configuration values provided by the app via configuration files or via environment variables as specified by `config/custom-environment-variables.json`, `config` processes my defaults as given to `CONFIG.util.extendDeep(...)` & `CONFIG.util.setModuleDefaults(...)` just fine.  The moment I add environment variables or configuration files that would override the module defaults, I get the same error as @rpaterson.

matthewadams:
Looks like it was introduced in version `1.20.2` (see output below).  If I roll back my project to `config@1.20.1`, everything's jim-dandy.

```
$ for v in 1.21.0 1.20.4 1.20.3 1.20.2 1.20.1 1.20.0; do echo $v ; npm i --save config@$v ; node index.js ; done
1.21.0
npm WARN package.json test@1.0.0 No description
npm WARN package.json test@1.0.0 No repository field.
npm WARN package.json test@1.0.0 No README data
config@1.21.0 node_modules/config
└── json5@0.4.0
/Users/matthew/temp/config/test/node_modules/config/lib/config.js:1434
[...]

PR #383, opened 2017-02-08
fix for #329

dcolens:
see https://github.com/lorenwest/node-config/issues/329#issuecomment-278300660

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1464,7 +1464,7 @@ util.extendDeep = function(mergeInto) {
       }
 
       // Copy property descriptor otherwise, preserving accessors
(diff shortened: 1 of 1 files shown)


markstos:
Not only this does PR not pass the test suite, but it changes the code to call `.get` on the return value of `getOwnPropertyDescriptor` which [is documented](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor) to return `undefined` sometimes. This patch would then result on a fatal attempt to call `.get` on an undefined value, throwing an exception.

You are welcome re-submit if you've got a better idea that passes the test suite and avoids this issue as well. 

Closing this PR for now.

PR #384, opened 2017-02-13
handles case where Object.getOwnPropertyDescriptor is undefined (Fixes #329)

### LINKED ISSUE #329: "Cannot redefine property" error using util.extendDeep to extend a config

dcolens:
handles case where Object.getOwnPropertyDescriptor is undefined.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1464,7 +1464,7 @@ util.extendDeep = function(mergeInto) {
       }
 
       // Copy property descriptor otherwise, preserving accessors
-      else if (Object.getOwnPropertyDescriptor(Object(mergeFrom), prop).get){
+      else if (Object.getOwnPropertyDescriptor(Object(mergeFrom), prop) && Object.getOwnPropertyDescriptor(Object(mergeFrom), prop).get){
           Object.defineProperty(mergeInto, prop, Object.getOwnPropertyDescriptor(Object(mergeFrom), prop));
       } else {
           mergeInto[prop] = mergeFrom[prop];
--- a/lib/config.js
+++ b/lib/config.js
@@ -1464,7 +1464,7 @@ util.extendDeep = function(mergeInto) {
       }
 
       // Copy property descriptor otherwise, preserving accessors
-      else if (Object.getOwnPropertyDescriptor(Object(mergeFrom), prop)){
+      else if (Object.getOwnPropertyDescriptor(Object(mergeFrom), prop).get){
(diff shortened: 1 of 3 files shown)


markstos:
Thanks for helping to improve `node-config`. To merge this, we also need a new test for the test suite that fails before this patch is made, but passes it after applies. 

From reading the [docs on getOwnPropertyDescriptor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor), it appears to cover a case where a property descriptor is defined, but the getter on the property descriptor is not defined. 

This change would mirror the change I made in db2e1357565bba3665a257e9da7187b9e668cbb9 which also checked both for a property descriptor and a getter. 

Looking at the history of this specific area of code, I see that db2e1357565bba3665a257e9da7187b9e668cbb9 introduced the same kind of bug during a refactor.  Then 3b221d756ba198b6c74a37ae9f5556b2a72909f5 tried to fix it but also failed to check that the getter was defined. 

This troubled history of flaws patches illustrates why it's important to have tests for the all the interesting cases.

dcolens:
Including the test suggested in Pull request https://github.com/lorenwest/node-config/pull/333 in this pull request.

markstos:
I'm closing this because I think the case it improves doesn't exist in any of the version of Node.js that we currently support now in 2021. If there's a case where this can be triggered in a supported version of Node, please explain.

PR #385, opened 2017-02-14
Update config.js with correctly cased type def.

ScionOfBytes:
Some IDEs and other tools incorrectly state that argument of type `String` is not assignable to param of type `STRING`. This fixes those sort of issues.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1633,8 +1633,8 @@ util.getCmdLineArg = function (searchFor) {
  * </p>
  *
  * @method getEnv
- * @param varName {STRING} The environment variable name
- * @return value {String} The value of the environment variable
+ * @param varName {String} The environment variable name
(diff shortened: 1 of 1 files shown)


markstos:
I reviewed this. Looks like just code-comment changes which will be neutral for most people, but important to people that use IDEs like @ScionOfBytes does.   

@ScionOfBytes What's an example of an IDE or tool that cares about this difference?

ScionOfBytes:
@markstos It keeps telling me that I can't pass values of type `String` to a function that takes params of type `STRING`. Similarly since {Mixed} isn't a part of the jsDoc spec, WebStorm kept telling me I can't pass this or that to something of type `Mixed`.

markstos:
It's good to know it was Webstorm-- that's a popular IDE!

PR #392, opened 2017-03-10
Flag for no custom env

lecstor:
I would like to be able to ensure that ENV vars do not override my test config when running tests in CI.I could not see how that would currently be possible (in a simple way). Would something like this be an acceptable option?

(needs some cleanup too but I'm out of time right now..)
..and it should probably stop NODE_CONFIG too

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -538,7 +538,7 @@ util.makeImmutable = function(object, property, value) {
   for (var i = 0; i < properties.length; i++) {
     var propertyName = properties[i],
         value = object[propertyName];
-    
+
     if (!(value instanceof RawConfig)) {
       Object.defineProperty(object, propertyName, {
         value: value,
@@ -746,8 +746,10 @@ util.loadFileConfigs = function(configDir) {
   }
 
   // Override with environment variables if there is a custom-environment-variables.EXT mapping file
-  var customEnvVars = util.getCustomEnvVars(CONFIG_DIR, extNames);
-  util.extendDeep(config, customEnvVars);
+  if (process.env.NO_CUSTOM_ENV !== 'true') {
+    var customEnvVars = util.getCustomEnvVars(CONFIG_DIR, extNames);
+    util.extendDeep(config, customEnvVars);
+  }
 
   // Extend the original config with the contents of runtime.json (backwards compatibility)
   var runtimeJson = util.parseFile(RUNTIME_JSON_FILENAME) || {};
--- a/test/11-config/custom-environment-variables.json
+++ b/test/11-config/custom-environment-variables.json
@@ -1,0 +1,8 @@
+{
+  // With a comment
+  "customEnvironmentVariables": {
+    "mappedBy": {
+      "json": "CUSTOM_JSON_ENVIRONMENT_VAR"
+    }
+  }
+}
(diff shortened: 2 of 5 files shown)


markstos:
May I ask which CI environment you are in and what the troublesome variable is? 

We have a lot of users and this hasn't come up before.

What about just deleting the problematic environment variables in the CI environment before loading node-config?

lorenwest:
Here's a workaround that doesn't require introducing code and another config configuration variable:

```
var hold = process.env;
process.env = null;
process.env = hold;
```
Do that in your bootstrap script - the first module that does a ```require('config')``` - and you won't have to worry about those pesky environment variables getting in the way.

lecstor:
Fair question, but as it turns out the issue that instigated this wasn't caused by CI or ENV. We're using Buildkite, but that's irrelevant anyway.
My issue was that it looked like it could have been caused by incorrect config, which could have been caused by env vars (though it shouldn't..), but it wasn't something I could definitely rule out.
quoting the boss: `tests should not use ENV, should have full control over config`
I'm pushing for more use of node-config so this just didn't help that.

lorenwest:
Good thought, although a test that relies on `NO_CUSTOM_ENV` environment variable can't really pass the `tests should not use ENV` requirement. The solution above would assure your code doesn't use environment variables, although it would exclude them from tests as well as production.

It's kind of a catch-22.

lecstor:
@lorenwest mm, 'k thanks. not sure how clean or simple that will be using jest, but will look into it. Was hoping for a nice solution. We have very many services and many different engineers creating new ones all the time. Will see what I can figure out. thanks.

lecstor:
8) yeh, totally agree..

markstos:
In my `gulpfile` that runs the tests, I explictly set ``process.env.NODE_ENV`, `process.env.NODE_APP_INSTANCE` and `process.env.MONGODB`-- I want there to be no way that the tests could run with the wrong configuration or connect to a non-test database.

The solution @lorenwest provided has essentially the same effect.

If the issue is potential confusion about what's going on with all the environment variables, I don't think adding another environment will help that.

I recommend explicitly setting (or un-setting!) the critical environments in your test environment.

lecstor:
yes, agree again. But it's the custom env vars that are the main issue. I do set NODE_ENV, we don't use NODE_APP_INSTANCE, but we do use lots of server config env vars common across many services which I merge into my apps' configs with custom env vars. It's annoying because in reality these env vars should/will never be set in the test env, but having said that we have seen assorted issues with things not being cleared out from previous runs of different service tests, so it's just something I'd like to be able to say, "nope, it wouldn't read them even if they were there".

lecstor:
would a command line option be any more palatable?

markstos:
@lorenwest is the primary maintainer, but my opinion is to not add a new option for this based on a single request. We have over 350,000 downloads per month now and new issues opened almost everyday, with perhaps a surprising number of one-off requests for new options and features to be added that are only ever requested by a single person. 

You can imagine the result if we added every feature that was requested for a single person or small minority. 

In this case, it seems like there's a fairly simple and intuitive workaround to just clear the environment in the test environment.  Yes, there's a risk that you could forget to do this, but there's also a risk you could forget to set `NO_CUSTOM_ENV` or add a new command line flag.

I'm going to close this request now but we'll consider it further if there's interest from a number of other people.

lorenwest:
Personally, I prefer the "clean up your house" approach. Control the environment your servers are running within.

If tests are failing because the environment isn't well controlled, those tests have served their purpose.

Issue #226, opened 2015-06-14
Find a solution to the multiple-instance problem for submodule defaults


lorenwest:
When multiple instances of node-config are loaded, only one instance will get module defaults inserted on `setModuleDefaults()`.  Other configurations will be duplicated, so it's only an issue for submodule defaults.

This is related to issue #225 because defaults loaded under submodule /config directories will have the same problem as defaults loaded using `setModuleDefaults()`.

Some ways this could be addressed:

    * Assuring copies are identical (hard)
    * Having all copies work on a single global config (messy)
    * Assuring only one instance is loaded (not sure this is possible)
    * Others?

If we were to resolve this issue, which one makes most sense?

markstos:
Is it fair to interpret this issue as not being just about `setModuleDefaults()` but more generally the issue `require('config')` is cached and thus represents only one configuration at a time?

lorenwest:
The feature of having a single config across all modules was implemented as cacheing the first ```require('config')```. There are many subtle side-effects of this implementation, one of which being the issue in this thread.

This thread is specifically about improving the side effect of ```setModuleDefaults()```. I don't think it's right to characterize it more generally to include all side effects of the ```require('config')``` cache. That could be a topic of another thread.

jdmarshall:
The solution for this probably involves pulling configSources out and storing them somewhere else, but then you also have to worry about forward and backward compatibility with the sources, since now multiple versions of node-config will be reading and writing them.  

Personally I think the problem is in the opposite direction.  If modules are 'publishing' their configurations where the rest of the app can read them, that not only makes node-config a singleton (which it didn't need to be, and should be fixed in 4.0), but also global shared state. 

It's none of the rest of the app's goddamned business what my module's configuration is. Nobody should be looking at it except the module itself. If you're doing anything else you've established one of the worst flavors of covert channel and you need to reconsider your (lack of) architecture.  All I need as a function buried somewhere deep in node_modules is for the application to be able to feed me overrides, NOT the other way around. For that, you need to be able to share the results of the initial loadConfig call of the main application with any duplicate copies. That should be doable with a global.

markstos:
To frame this problem another way, isn't the same idea that there can be only be one "config" singleton?

So, if you are designing a module that uses node-config and may be a dependency of another module that uses node-config, then that's not safe: The first loads node-config will load the config and become the singleton. 

Related discussion is here: https://github.com/node-config/node-config/discussions/819

[...]

PR #393, opened 2017-03-13
Allow consumers of Config.prototype.get to specify a default value

konradjurk:
Currently, if I call `Config.prototype.get('var')`, and `var` cannot be resolved, the module throws an error. This is cool, but I'd like to be able to optionally specify a defaultValue as the second positional argument to avoid the exception or `Config.prototype.has()` roundtrip in my application.

I'd like to provide tests, but I don't understand the structure of the tests directory. What is the meaning of all the '3-config', '4-config', '5-config' ,..., directories? Could someone please point me to the right location? :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .project
 node_modules
 npm-debug.log
+.idea
--- a/lib/config.js
+++ b/lib/config.js
@@ -167,18 +167,24 @@ var getImpl= function(object, property) {
  *
  * @method get
  * @param property {string} - The configuration property to get. Can include '.' sub-properties.
+ * @param [defaultValue] {*} - Return this value if the property doesn't exist.
  * @return value {*} - The property value
  */
-Config.prototype.get = function(property) {
+Config.prototype.get = function(property, defaultValue) {
   if(property === null || property === undefined){
     throw new Error("Calling config.get with null or undefined argument");
   }
   var t = this,
       value = getImpl(t, property);
 
-  // Produce an exception if the property doesn't exist
   if (value === undefined) {
-    throw new Error('Configuration property "' + property + '" is not defined');
+    // Produce an exception if the property doesn't exist and the consumer didn't specify a default
+    if(!defaultValue) {
+      throw new Error('Configuration property "' + property + '" is not defined');
+    }
+
+    // continue to work with the given defaultValue
+    value = defaultValue;
   }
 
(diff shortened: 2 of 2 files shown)


markstos:
Closing as a duplicate of #381, which was found a quick search for "get" and default". 

If you want an exception to be thrown and don't want to use `has`, you can set a default value in the config system. You can even set the default value to `null` if you want. Only `undefined` throws an error.

konradjurk:
The whole point of this is to work with default values on the application level instead of some config file. The change as proposed here and in #381 is not breaking any compatibility and is completely optional, so the user can decide whether he wants to use the json file for default values or rather pass it to the get() call.

Lodash offers the same functionality in _.get(): https://lodash.com/docs/4.17.4#get

Please consider to accept #381.

markstos:
The whole point of `node-config` is to promote that configuration-- like default values-- should be in config files, not in code. Adding this change would result in projects that store their configuration partly in config files and partly in the code, which is the kind of problem the project was created to solve. 

If you want default values, that's what `config/default.js` is for.  Put your default values there. 

If you feel strongly about the usefulness of this feature, you are welcome to fork the project.

konradjurk:
@markstos Okay I think I see your point of view and I agree, partly. However, having every default value in a file doesn't make sense in _every single_ case. For example, I am wiring up a few node microservices which share an npm module for common stuff. This module uses your config project to resolve some config variables. I want the variables do have defaults. How am I supposed to do that without #381? I might end up with dozens of projects and every single one has to have a json file with default values for the shared module. That doesn't sound like a great idea, does it?

markstos:
I think the real issue is that we could use better support for "sub-module defaults". This issue is discussed more in existing tickets:

  https://github.com/lorenwest/node-config/issues/225
  https://github.com/lorenwest/node-config/issues/226

The issue affects projects like yours where `node-config` may be used both a top-level project and also be a dependency. Yet, currently `node-config` expects configuration to exist in a single top-level `config` directory, instead of looking for a different configuration for the sub-module. 

If that were better addressed, you could use a wrapper for the third-party NPM module which uses `node-config`. Your wrapper module could set the default values you want in a single config file of your own. This defaults could in a central place, separate from ./config directories of various projects which may use the third-party NPM module

konradjurk:
Okay, looks promising but it is already 2 years old. I'll go with my forked version. Thanks anyways.

PR #401, opened 2017-03-27
extendDeep test improvements

IvanVergiliev:
The result from extendDeep was never verified to be the same as the
expected value. This is informative of the behaviour of extendDeep since
there are only two tests that extend with more than one object and
neither of them verified the result before.


lorenwest:
Thanks for the added tests!

Issue #408, opened 2017-04-12
Reduce package size


evilebottnawi:
Package contains unnecessary directories and files - `tools`, `.travis.yml` and others.

lorenwest:
These comments are best expressed as pull requests

PR #409, opened 2017-04-13
Chore: reduce package size.

evilebottnawi:
Ref: https://github.com/lorenwest/node-config/issues/408
/cc @lorenwest


lorenwest:
Thanks for the cleanup

Issue #410, opened 2017-04-14
Bug: Overriding one Regex with another fails


XadillaX:
default:

```javascript
{
    foo: /foo/;
}
```

local:

```javascript
{
    foo: /bar/;
}
```

the `foo` won't be overwritten by local.

it ran into https://github.com/lorenwest/node-config/blob/4e1db3afb174878cf5920a38107a6f5e86b745b5/lib/config.js#L1469.

markstos:
Thanks for the report. It would also be helpful to have a failing test contributed to the test suite as well as a patch.

You could be the next one to have your photo added to the Contributors section: https://github.com/lorenwest/node-config/blob/master/README.md#contributors

XadillaX:
PR committed.

XadillaX:
I think this issue could be closed now.

PR #411, opened 2017-04-24
Fix the bug that merge from RegExp

XadillaX:
refer issue #410

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,2 @@
+[*.js]
+indent_size = 2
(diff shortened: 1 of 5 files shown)


markstos:
@lorenwest, this looks good to me.

lorenwest:
Thanks for the fix!

XadillaX:
Hi @lorenwest, I'm just here to ask when to release the new version? I'm waiting to use it.

lorenwest:
Published today @2.26.1 - thanks for the reminder

jdmarshall:
I think this PR likely reintroduced #197 due to a bad else if call.

lorenwest:
#197 was fixed by #198 which included a test to prevent the kind of regression you're suggesting.  Could you add a failing test that describes the issue you're seeing, and would pass once the regression is resolved?

PR #413, opened 2017-04-26
Added optional parameter to util.loadFileConfigs() method to override…

sakiv:
Util.loadFileConfigs() allows to read the configuration files from a different folder. It would be great to have another parameter to override APP_INSTANCE to read config files.

```
util.loadFileConfigs = function(configDir, instance) {
...
...
APP_INSTANCE = instance || util.initParam('NODE_APP_INSTANCE');
...
...
}
```

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -633,7 +633,7 @@ util.getConfigSources = function() {
  * @method loadFileConfigs
  * @return config {Object} The configuration object
  */
-util.loadFileConfigs = function(options) {
+util.loadFileConfigs = function(configDir, options) {
 
   // Initialize
   var t = this,
@@ -643,7 +643,7 @@ util.loadFileConfigs = function(options) {
 
   // Initialize parameters from command line, environment, or default
   NODE_ENV = options.env || util.initParam('NODE_ENV', 'development');
-  CONFIG_DIR = options.configDir || util.initParam('NODE_CONFIG_DIR', Path.join( process.cwd(), 'config') );
+  CONFIG_DIR = configDir || util.initParam('NODE_CONFIG_DIR', Path.join( process.cwd(), 'config') );
   if (CONFIG_DIR.indexOf('.') === 0) {
     CONFIG_DIR = Path.join(process.cwd() , CONFIG_DIR);
   }
@@ -713,7 +713,7 @@ util.loadFileConfigs = function(options) {
 
   // Override configurations from the $NODE_CONFIG environment variable
   // NODE_CONFIG only applies to the base config
-  if (!options.configDir) {
+  if (!configDir) {
     var envConfig = {};
     if (process.env.NODE_CONFIG) {
       try {
--- a/lib/config.js
+++ b/lib/config.js
@@ -633,7 +633,7 @@ util.getConfigSources = function() {
  * @method loadFileConfigs
  * @return config {Object} The configuration object
  */
-util.loadFileConfigs = function(configDir) {
+util.loadFileConfigs = function(configDir, instance) {
 
   // Initialize
   var t = this,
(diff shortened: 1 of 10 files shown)


lorenwest:
If you want to load configuration files from a different folder, why not use [$NODE_CONFIG_DIR](https://github.com/lorenwest/node-config/wiki/Environment-Variables#node_config_dir)?

sakiv:
Hi Loren,

Definitely we can use $NODE_CONFIG_DIR but to read configuration for different instances in same process, it would be great to pass in the instance parameter and get the config objects for individual instances.

In my use case, I have built a job scheduler which can have different configuration files for each job. I am using App_Instance to identify each job. 

I am not sure if I was able to articulate the need but it would be great to pass in instance as parameter to get the configuration object for all instances at one place.

sakiv:
@lorenwest, as @markstos suggested, it would be great to have a second argument that is an options object rather than just an instance value. The options object could contain an instance key as well as others.

I have updated the code branch with the same. Please let us know your views.

lib/config.js
```diff
@@ -633,25 +633,27 @@ util.getConfigSources = function() {
...
   // Initialize
   var t = this,
       config = {};
 
+  options = options || {};
+
   // Initialize parameters from command line, environment, or default
-  NODE_ENV = util.initParam('NODE_ENV', 'development');
+  NODE_ENV = options.env || util.initParam('NODE_ENV', 'development');
   CONFIG_DIR = configDir || util.initParam('NODE_CONFIG_DIR', Path.join( process.cwd(), 'config') );
```
lorenwest: What's the reasoning behind options for some parameters and not others?

lorenwest:
It appears this changes the functionality of a utility method, and not the functionality of node-config at all.

I can see how you could use this utility function to specify different parameters and return a different config object, but that doesn't allow the jobs you're running to be able to use `var config = require('config')` to get these different configurations.

How are these jobs you're running supposed to get their config instance?

lorenwest:
As important, how are dependent modules of your jobs supposed to get the right config? Does using node-config in this way prevent you from using dependent modules that also use node-config?

There's an ecosystem of modules that use node-config, and I'm not sure this is a good feature to add if it makes those sub-modules unusable. 

If I understand this correctly, while this may be good functionality for your use case, it may be best as a fork than a merge. We have a [wiki page](https://github.com/lorenwest/node-config/wiki/Plugins) to advertise useful additions for specific use cases.

sakiv:
Hi Loren,

Thanks for the reply and giving your thought on this. 

To answer your first question, each of my job runs in a different process and when trigger, an instance name is assigned to them. This resolves into picking the correct configuration files for each job. Even sub-modules will be able to resolve to correct config.

As you correctly mentioned, with options people will be able to use this as a utility function with different parameters. Keeping options as optional parameter will not break any existing functionality for sub-modules.

lorenwest:
Your use case sounds reasonable - kicking off a sub-process with different instance IDs, and it's a reasonable request to know what the config would look like before calling the job.

This is currently supported via altering `process.env.*` before/after calling `loadFileConfigs()`.

Give that a try and see if it fulfills your needs. I dislike adding code when the functionality is already present.

lorenwest:
Not sure what's going on in this pull request. Maybe you need to rebase your fork before issuing a pull request to this library. Far too many files changed to approve this as-is.

sakiv:
I have forked out from your repository as this pull request is not getting merged into master.

—
Thanks,
Vikas Aggarwal
Sent with Airmail


On December 4, 2017 at 12:20:12 AM, Loren West (notifications@github.com<mailto:notifications@github.com>) wrote:

Not sure what's going on in this pull request. Maybe you need to rebase your fork before issuing a pull request to this library. Far too many files changed to approve this as-is.

PR #340, opened 2016-08-05
Issue #338 Adding raw wrapper to prevent object modification in config

patrickpilch:
Looking for feedback on whether this is an acceptable solution that will cover all the scenarios. (e.g. does `util.cloneDeep` have relevance to this?)

**TODOS**

1. Improve test
2. Documentation in the Wiki

PR #426, opened 2017-07-22
Add .raw method

### LINKED ISSUE #223: Add a method for outputting a plain JS object (PR welcome)
Feature request:  

I think it would really benefit this module to formalize the output of a plain javascript object. I know the little one-liner `JSON.parse(JSON.stringify())` but boy, that feels like an opaque weird thing to have in my app code. The need for POJOs is frequent, for instance, using the hapijs server if I attempt to bring any `node-config` object into a config, hapi will fail on the schema validations because of all the additional methods.

Other libraries, such as immutable, that do large amounts of manipulation on objects offer a `toJS` method, which makes a lot of sense to me. As a stop-gap that method could just do the parse/stringify trick, but I think ideally it would access the hidden internal variables and output those directly.

Going even further, I'd make the case that _any_ invocation of `.get()` should actually output a POJO. It feels highly unnatural to me to be carrying around all this baggage when I've clearly made the choice to only grab a portion of my config.

Kikobeats:
Fix #223 

I normally need to acces to the raw `config` file, this is the trick that I do:

```js
const config = JSON.parse(JSON.stringify(require('config')))
```

I simply put into the codebase for do that transparently.

Out of the box works, but not sure if inside of the core it will work, but I want to init this PR 😄

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -155,6 +155,9 @@ var getImpl= function(object, property) {
   return getImpl(value, elems.slice(1));
 };
 
+Config.prototype.raw = function() {
+  return JSON.parse(JSON.stringify(this)))
+}
 
 /**
  * <p>Get a configuration value</p>


lorenwest:
This would be up for consideration if

1) It includes a test
2) You've run the test suite and fixed the syntax error
3) You add documentation about this functionality to the wiki

markstos:
Closing because "raw" support was added in https://github.com/lorenwest/node-config/pull/340

PR #435, opened 2017-09-27
feat(util) - move util methods to separate file. 

Kinjeiro:
Case: use loadFileConfigs in default.js to load parent config from other folder BEFORE new Config() occurs


package.json
```diff
@@ -45,6 +45,6 @@
     "node": ">= 4.0.0"
   },
   "scripts": {
-    "test": "./node_modules/vows/bin/vows test/*.js --spec"
+    "test": "vows test/*.js --spec"
```
markstos, line 48: Why change this? Before, it was self-contained.   Doesn't this change create a new dependency on a global installation of "vows"?

markstos:
This is a huge diff, which makes it hard to see what the intended change is, compared the restructuring that  was done. It would be easier to review if there was at least 2 commits: One that refactors the code but doesn't change the logic, and another that makes the logic change. 

There is also no test coverage for the new functionality. Could you provide a code example that shows what didn't work before the change that would after the change?

lorenwest:
I'm confused to the need for this change. Can you point me to the discussion that suggested it?

package.json
Kinjeiro: There is no need to install vows globaly. When run "npm run test" node will search "vows" comand in <project>/node_modules/.bin/vows.sh (or vows.cmd for Windows) first.

Kinjeiro:
@markstos Thank you. I will rollback and seperate this commit to few small commits. All existed tests are passed.

@lorenwest My point is: if you will use "import config from 'config'", there is new Config() initialize, which run loadFileConfigs for current local directory. But i want to load config from other folder first, before local config. But i can't do this, because loadFileConfigs has been already run.

Example: there are 2 repositories
myFirstRepo
-- config
-----default.js
-----development.js
mySecondRepo
--config
----default.js
----development.js
--package.json (which has in dependecies on myFirstRepo)

So i want to load in mySecondRepo "node_modules/myFirstRepo/config" first, and then mySecondRepo/config. 
I want to write:

**_mySecondRepo/config/default.js_**
```javascript
const {
  extendDeep,
  loadFileConfigs
} = require('config/lib/util');

const { inNodeModules } = require('./path-utils');

module.exports = extendDeep(
  // parent config
  loadFileConfigs(path.join(process.cwd(), 'node_modules/myFirstRepo/config')),
  {
    common: {
      features: {
        i18n: {
          ns: ['core', 'portal']
        }
      }
    },
    client: {
    },
    server: {
    }
  }
);
```

Then 
**_mySecondRepo/src/server/index.js_**
```javascript
import config from 'config';
// there is already merged config from myFirstRepo and mySecondRepo
```

lorenwest:
This is a major change with no public discussion.

Too many changes in this branch would disrupt existing installs, and clashes with node-config philosophies. I suggest we discuss the issues in public forum before considering a change in behavior, let a lone a particular implementation of that change.

Alternately you could publish this as a forked repo.

PR #438, opened 2017-10-15
Add method to output plain JS object

### LINKED ISSUE #223: Add a method for outputting a plain JS object (PR welcome)

willsoto:
Closes #223

I will be happy to add this to the Wiki if this is acceptable.

PS: I saw #426, but it looks abandoned, so I did it here taking into consideration [this comment](https://github.com/lorenwest/node-config/issues/223#issuecomment-112951557)

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -538,7 +538,7 @@ util.makeImmutable = function(object, property, value) {
   for (var i = 0; i < properties.length; i++) {
     var propertyName = properties[i],
         value = object[propertyName];
-    
+
     if (!(value instanceof RawConfig)) {
       Object.defineProperty(object, propertyName, {
         value: value,
@@ -867,7 +867,7 @@ util.parseFile = function(fullFilename) {
     if (extension === 'js') {
       // Use the built-in parser for .js files
       configObject = require(fullFilename);
-    } 
+    }
     else if (extension === 'ts') {
       require('ts-node').register({
         lazy: true,
@@ -1668,6 +1668,16 @@ util.getRegExpFlags = function (re) {
   re.ignoreCase && (flags += 'i');
   re.multiline && (flags += 'm');
   return flags;
+};
+
+/**
+ * Returns a new deep copy of the current config object, or any part of the config if provided.
+ *
+ * @param {Object} config The part of the config to copy and serialize. Omit this argument to return the entire config.
+ * @returns {Object} The cloned config or part of the config
+ */
+util.toObject = function(config) {
+  return JSON.parse(JSON.stringify(config || this));
 };
 
 // Run strictness checks on NODE_ENV and NODE_APP_INSTANCE and throw an error if there's a problem.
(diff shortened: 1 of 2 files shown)


willsoto:
Looks like the CI error has something to do with Node 4? 

https://travis-ci.org/lorenwest/node-config/jobs/288266571#L462

lorenwest:
This code looks good - thanks for the addition. It's an old thread, and I'm glad someone who needed the functionality added it.

The failure appears to be a NodeJS versioning problem. I'll see if I can get it working again.

lorenwest:
Got it working. Rebase and re-submit, and I'll get it merged.

willsoto:
@lorenwest Thanks for that! I'll take of that this evening and ping you.

willsoto:
@lorenwest That did it. Rebased and CI is green.

willsoto:
I will also have some time tonight to add this to the Wiki (if that is still desirable). Would [this](https://github.com/lorenwest/node-config/wiki/Using-Config-Utilities) be the right page for it?

lorenwest:
Yes, that would be the perfect page for it.  Thanks for the update!

lorenwest:
Published changes today on v1.27.0

markstos:
@willsoto Thanks for the contribution. Did you also add docs for `.toObject()?` I don't see it mentioned in the README or on the wiki.

willsoto:
https://github.com/lorenwest/node-config/wiki/Using-Config-Utilities#toobjectconfig

markstos:
Thanks.   Perhaps we should also add an FAQ page  as well. Someone just ran
into this and posted on StackOverflow:

https://stackoverflow.com/questions/48867065/typeerror-key-must-be-a-buffer-when-sending-an-email-with-aws-sdk-in-node-js/49032379#49032379

Essentially, they want to know "Is node-config missing with my data? Why?
How do I fix it?".

The function docs for `toObject()` are only useful once you've realized
precisely that you need such a method. As seen on #223, there have been
about 10 different related issues opened.

Maybe we just need to be more up front to explain that node-config is going
to add some methods to objects it turns, along with the solution for
getting rid of the extra methods if you don't want them.

PR #447, opened 2017-11-14
Add failing native promise test

scottmas:
Node has had native promise support for a long time. However, defining a native promise in your config results in a value that cannot be successfully subscribed to. I suspect some sort of weird binding is occurring somewhere inside of config, which causes the `this` of the promise to be incorrect.

### DIFF
--- a/test/3-config/default.js
+++ b/test/3-config/default.js
@@ -5,7 +5,7 @@ var config = {
   siteTitle : 'Site title',
   latitude  : 1,
   longitude : 2,
-
+  nativePromise: Promise.resolve('A native promise value')
 };
 
 // Set up a default value which refers to another value.
--- a/test/3-deferred-configs.js
+++ b/test/3-deferred-configs.js
@@ -49,6 +49,12 @@ vows.describe('Tests for deferred values - JavaScript').addBatch({
     "second defer function return original value." : function () {
       assert.equal(CONFIG.original.deferredOriginal, undefined);
     },
+
+    "native promise can be subscribed to" : function () {
+      return CONFIG.nativePromise.then(val => {
+        assert.equal(val, 'A native promise value')
+      })
+    }
   }
 })
 .export(module);


markstos:
I confirmed that promises are supported using the `raw()` feature already present and documented here:

https://github.com/lorenwest/node-config/wiki/Configuration-Files#javascript-module---js

I think this feature is unfortunately harder to find then it should be. Perhaps we should split out the JavaScript support on to it's own page on the wiki, now that we have both `defer()` and `raw()` support which are specific to JavaScript. 

I'll open an issue for that, #448

scottmas:
That would be a good idea to add better documentation for that. It's good to know that I can solve this issue by wrapping my promise in raw. That said, while I don't know the details of how you make the config object prototypes immutable, it does seem reasonable that a promise should be resolvable. It's not a mutable value per se. It's just an object that depends on its `this` context when you subscribe to it. 

So could there be a roadmap for supporting native promises without wrapping them in `raw`?

markstos:
@scottmas you are welcome to look into it. I think you'll find that every config value normally gets wrapped into it's own object to make the module work, which is why `raw()` exists. We also need to support the `defer()` case.  I suspect it will be a challenge to have promises continue to work as promises while also meeting the needs currently served by the custom `node-config` objects that values get wrapped in.  Your first stop should to look at how `raw()` and `defer()` are implemented and then now `node-config` handles config values internally. 

There's also the issue that `node-config` is always loaded synchronously, while promises by their nature may be resolved async.

lorenwest:
We've done well with the custom `defer` logic, and I believe the next step in the evolution of `node-config` is native Promise support.

Node-config is a synchronous library. You can `get()` configurations at module load time and count on their stability for the life of the application. This stability is a core differentiator for `node-config`, yet has prevented it from natively representing configurations with underlying asynchronous operations.

Promises are stable. You can `get()` a reference to them at module load, and they have a stable API for referencing asynchronous data from that reference. They do mutate, but only once, so we would need custom logic for Promises that bend this mutation restriction.

With that, I'm requesting community assistance to help build custom Promise support into this library, with documentation on it's direct usage as well as async/await usage.

lorenwest:
This feature will not work with prior NodeJS versions without requiring bluebird or other Promise library. We don't want to add bluebird as a dependency, so this will break installs on versions of NodeJS without native Promise support.

As such, we will bump the major version to `2` with this feature, and the `2` version will only support versions of NodeJS with native Promise support.

markstos:
@lorenwest,

>  so this will break installs on versions of NodeJS without native Promise support.

> As such, we will bump the major version to 2 with this feature, and the 2 version will only support versions of NodeJS with native Promise support.

We already require at least Node.js >= 4 (at least, that's all we test on Travis). yet according to http://node.green/ basic promise support was introduced in Node.js 0.12 and is already there in Node.js 4.

So, unless we need  Promise support beyond what's already in Node.js 4, it's not clear that we need drop support for any versions of Node.js that we currently support.

scottmas:
Closing in favor of https://github.com/lorenwest/node-config/issues/449

PR #456, opened 2017-12-22
Fixed 'hostname' value selection when there is no environment variable for that

wmangelardo:
Hi,

We are facing some issues in some distros that do not have HOSTNAME (or HOST) environment variable setted. In this case, we are receiving `undefined` result when trying to `config.util.getEnv('HOSTNAME')`.

This PR is a suggestion to try to do what says in the comments (lib/config.js:662)

> Determine the host name from the OS module, $HOST, or $HOSTNAME

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -664,9 +664,6 @@ util.loadFileConfigs = function(configDir) {
   try {
     var hostName = HOST || HOSTNAME;
 
-    // Store the hostname that won.
-    env.HOSTNAME = hostName;
-
     if (!hostName) {
         var OS = require('os');
         hostName = OS.hostname();
@@ -674,6 +671,9 @@ util.loadFileConfigs = function(configDir) {
   } catch (e) {
     hostName = '';
   }
+
+  // Store the hostname that won.
+  env.HOSTNAME = hostName;
 
   // Read each file in turn
   var baseNames = ['default', NODE_ENV];
--- a/test/13-node_env-hostname.js
+++ b/test/13-node_env-hostname.js
@@ -1,0 +1,54 @@
+var requireUncached = require('./_utils/requireUncached');
+
+'use strict';
+
+var NODE_CONFIG_DIR = __dirname + '/12-config'
+
+// Dependencies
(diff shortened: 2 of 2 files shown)


markstos:
@lorenwest Why do we set values in the environment at all? Besides setting `env.HOSTNAME` I also see that `initParam` has the undocumented feature of setting the params into the environment as a side-effect.

If we want to continue with this behavior in `initParam`, it should be documented.

lorenwest:
Outside of tests, I'd be shocked if we're mutating environment variables. We don't have the (social) authority to do that.

A quick audit didn't show any place we're doing this. The code in this pull request is setting the `env` variable, which has module static scope. I don't see anything wrong with this pull request - it appears to be a healthy change.

markstos:
@lorenwest We set environment variables in these two places. The commits date to work you did in 2014:

https://github.com/lorenwest/node-config/blob/master/lib/config.js#L1615
https://github.com/lorenwest/node-config/blob/master/lib/config.js#L668

lorenwest:
Either I'm reading this wrong or you are. No matter how much I try to see how this sets an environment variable, I can't see it. What am I missing?

`env` is not `process.env`, it's a variable declared in module scope.

wmangelardo:
@markstos, I think what @lorenwest means is the lines that you cited refers to a variable declared at [line 28](https://github.com/lorenwest/node-config/blob/master/lib/config.js#L28) (module static variable).

There are no changes in `process.env` environment variables.

markstos:
@wmangelardo @lorenwest  Thank for the clarification about the scoping. Now that I've got my head on straight, I think this patch looks like an improvement.

The intent of the logic was to store which version of the hostname "won". The patch not only considers the HOST or HOSTNAME environment variables, but also `os.hostname()` if we ended up using that.

I think this just hasn't come up because I expect this edge case has rarely mattered.

lorenwest:
Pushed to npm config@1.29.0

Thanks for the contribution!

wmangelardo:
Thank you, @lorenwest and @markstos , for your attention!

Issue #442, opened 2017-10-23
Allow default values for config retrieval


BrainMaestro:
**Note: for support questions, please use [StackOverflow](https://stackoverflow.com/questions/tagged/node-config) and tag your question with `node-config`**. *This repository's issues are reserved for feature requests and bug reports.*

Before submitting a bug report, please search the issue tracker the wiki first. Many issues have already been discussed.

The wiki is located at: https://github.com/lorenwest/node-config/wiki

### I'm submitting a ... 
  - [  ] bug report
  - [x] feature request
  - [  ] support request or question => Please do not submit support request or questions here, see note at the top of this template.

### Problem
`config.get('property')` gets a property for a config (obviously). I would like a way to provide a default value if that property is missing or undefined. Currently, there are two ways of handling this:

1. `config.get('property') || 10`. This is easy and does not require any library modifications. However, it fails for falsy values like `false`, `''`, or `0` which could be valid configs. 
2. Using a more explicit `typeof` or strict equality check. This is more accurate, but it litters the code with unnecessary if statements. If there were 5 config retrievals, the check would be duplicated each time.

### Proposal
Allow the user to provide a default in case the value is undefined or null. This would look like `config.get('property', 10)`. Elegantly solves the problem with little duplication.

#### Prior Art
https://github.com/illuminate/config/blob/master/Repository.php#L47

I would be willing to implement this if it is accepted

lorenwest:
I'd support a pull request for this feature.

markstos:
This feature request was previously discussed and rejected in #393 and #381

I believe this feature is harmful.  The current design encourages that all default values should be in the config system, where they belong.

Adding the ability to set defaults in the code confuses where they belong. Should default values still be in `default.js`, or should they now be sprinkled throughout the code? Or some mix? 

Before my large project used `node-config`, we had a simpler config system, and a lot of default values sprinkled throughout the code base.  It was a pain to understand the configuration as a whole because it was spread across the entire codebase in adhoc locations. Sometimes this kind of "default" syntax gets used in multiple locations with different values, so there ends up being three different "default" values stored in three different locations. Bad. 

If you are tempted to use `config.get('property') || 10`, then `10` is a magic number, a constant that belongs in your config system, not in your code. Put in in `config/default.js`.

[...]

PR #462, opened 2018-01-26
Support optional default value in the get method

kureci:
and tests to cover the change

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -169,7 +169,9 @@ var getImpl= function(object, property) {
  * @param property {string} - The configuration property to get. Can include '.' sub-properties.
  * @return value {*} - The property value
  */
-Config.prototype.get = function(property) {
+Config.prototype.get = function(property, optionalDefault) {
+  var defaultValue = (typeof optionalDefault === 'undefined') ? undefined : optionalDefault;
+
   if(property === null || property === undefined){
     throw new Error("Calling config.get with null or undefined argument");
   }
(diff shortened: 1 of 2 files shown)


kureci:
This is my first fork/PR on an open source project, ever! So please be nice :)

markstos:
@kureci I appreciate you are getting involved with open source. 

In the future, I recommend searching the issue tracker and previous PRs before submitting one, because this request is a duplicate of a requests/PR that has already been submitted and rejected  a number of times. The most recent being #442  You can read more about the rationale, there. 

Although this PR was not accepted, the fact that you added test coverage is something that would increase the chances that a project will accept a PR from you in the future. It's also helpful to update the documentation when you add a new feature as well.

kureci:
I understand the rationale you've given in #442, and we've also had the same discussions with my work colleagues. However, there are also cases where you should be allowed to just accept a default if no config exists for it. These can be non-critical config parameters that is nice to have in configuration but not absolutely necessary. 

Allowing to have default values in code isn't the same as supporting the use of it. We should still discourage using the functionality in most cases, but still give people the option for rare cases. This is especially true for fast-paced environments where CI/CD is key; when you use configuration management tools like Puppet and it takes a bit longer to deploy configuration changes to environments (for various reasons), such restriction blocks application code to be deployed before configuration without breaking something (i.e. the exception thrown in this case).

Many people would discourage using `static` methods in most cases, but that doesn't mean you should remove the ability to create static methods in a language.

kureci:
That being said, I now see my issue. The module neatly allows the use of `default.json` file which can/should really be in your repo that overcomes the hurdle I described above. My problem was how I didn't want to have `default.json` in the repo, but now I see the benefit. 

Now I agree with you :)

markstos:
For clarity of *all* the config options, I recommend putting *every* config value in `config/default.js`, including optional values.

I don't understand how an environment is both "fast paced"  yet "takes longer to deploy changes". 

Since the feature add default values to `config.get()` doesn't exist, I know that my team isn't using it to sprinkle config values througout the code.  There is already Lodash `_.get()` if you want put default values in functions that for some reason don't seem to belong in central configuration.

kureci:
"fast-paced" !== "everything is deployed at the same time" ;)

PR #463, opened 2018-01-27
Add info about IntelliJ-plugin.

annitya:
A shameless plug from yours truly here.

The plugin provides references between usage of node-config and the config-files themselves.
This allows for code-completion, find-usages, rename-refactoring and the like.

While still in its infancy, I believe it should provide a little productivity boost already now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,10 @@ Running in this configuration, the `port` and `dbName` elements of `dbConfig`
 will come from the `default.json` file, and the `host` element will
 come from the `production.json` override file.
 
+**IntelliJ-plugin:**
+
+ Install `NodeConfig` from the IntelliJ [plugin-repository](https://plugins.jetbrains.com/plugin/10381-nodeconfig).
+
 Articles
 --------
 


markstos:
The screenshot shows on the plugin page is a bit weird:
https://plugins.jetbrains.com/plugin/10381-nodeconfig

First, I think it's most common to import `config` with the name `config`,
but `stats_db` is used here. Second, config values are usually accessed
with `config.get('some-name')`, but the autocomplete seems to promoting
direct access to the configuration property names, which is discouraged The
autocompletion that people are most likely to want-- `get()`-- does not
appear in the list.

whitefire:
I modeled the usage-pattern after that of our organization. Seems we've been doing it wrong...

No worries though, that can easily be remedied :)
Would you prefer I remove all support for direct-access all together or keep it as an alternative?

markstos:
Usually calling `config.get()` or `config.has()` is a much better option than direct access. You could continue to provide it if you want, but I think it could primarily serve to add confusion.

whitefire:
Best practice is the way to go. I will make adjustments! :)

annitya:
Well... it sure took me long enough, but version 2.0 is now awaiting approval from JetBrains.
I don't know if you're an IntelliJ man yourself, but the new version is available for testing [here](https://github.com/annitya/nodeconfig/blob/2.0/node-config.jar).

image

Hopefully I didn't lure too many developers to the dark side in the meantime. Old format-support is not only removed, I also added an inspection highlighting non-recommended usage for good measure.

lorenwest:
What do you think @markstos ? I'm appreciative of the work @annitya put into this, including your suggested changes.

markstos:
@annitya The testing link is 404. When it's working,I'll have some co-workers who use WebStorm test.

annitya:
Which link are you referring to?

markstos:
@annitya https://github.com/annitya/nodeconfig/blob/2.0/node-config.jar is 404 for me.

annitya:
Ah, now I understand what you mean. You needn't download the plugin from GitHub, just install it like you would any other plugin. Let me know of any improvements I can make, and of course if you find a bug! :)

markstos:
@annitya I had my co-worker install version 2.0.2 of the plugin. I saw that it completed the raw property names, but it not values within `config.get()` as seen in your screenshot. 

Regardless, @lorenwest I think the more appropriate place to add this would be on the "Community Utilties" wiki page, rather than in the README. Would you agree?

https://github.com/lorenwest/node-config/wiki/Community-Utilities

lorenwest:
Yes, add to community utilities page.

markstos:
@annitya I believe you edit the related wiki page directly, so I'm closing this PR.

PR #464, opened 2018-02-01
Support multiple config for diffrent branch in one process.

lanistor:
Add configrc.* file to read config directory;
configrc.* file can support multiple config for diffrent branch in one process.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1730,13 +1730,13 @@ util.toObject = function(config) {
  */
 util.getConfigDirFromRc = function(dir) {
   dir = typeof dir === 'undefined'? __dirname : dir;
-  let p = Path.resolve(dir, CONFIG_RC);
-  let conf = util.parseFile(p);
+  var p = Path.resolve(dir, CONFIG_RC);
+  var conf = util.parseFile(p);
   if(conf && conf.NODE_CONFIG_DIR) {
     CONFIG_RC_MERGE = conf.config;
     return conf.NODE_CONFIG_DIR || undefined;
   }else {
-    let ps = Path.parse(p);
+    var ps = Path.parse(p);
     if(ps.dir === ps.root) {
       return;
     }else {
--- a/lib/config.js
+++ b/lib/config.js
@@ -660,7 +660,7 @@ util.loadFileConfigs = function(configDir) {
   // Initialize configrc from configrc
   USE_CONFIG_RC = util.initParam('USE_CONFIG_RC', USE_CONFIG_RC);
   CONFIG_RC_DIR = util.initParam('CONFIG_RC_DIR', __dirname);
-  CONFIG_RC = util.initParam('CONFIG_RC', 'configrc');
+  CONFIG_RC = util.initParam('CONFIG_RC', 'configrc.json');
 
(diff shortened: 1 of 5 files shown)


lorenwest:
What is this solving? Was there a discussion of this that I missed?

lorenwest:
This breaks our Travis CI tests

vifird:
Sorry, it's because using `let`, i have repaired this problem.
`node-config` is a great library, we dependent on it deeply in our projects.
**Problem in usage**
But when we need deploy our project progresively, we found it cannot fit us well: we need to deploy multiple versions' code of same project in one process, then support a router to dispatch them, and each version need to have their own configs, then we found `node-config` cannot do it.
**Our solution**
To resolve this problem, we change `node-config`, add a file named `configrc.*` to read `NODE_CONFIG_DIR`: when set this environment, `node-config` will first finding `configrc.*` to read `NODE_CONFIG_DIR` in `__dirname`, when not found, it will recursively find the upper level directory, until reach root directory.
**Usage**
```
USE_CONFIG_RC=true CONFIG_RC=configrc.json5 node server
```
**Test**
In `./test/14-node-configrc.js`.

markstos:
So to make sure I understand your situation: You've got a big project, one where part of it might depend on a project that uses `node-config` and is at version 3, and elsewhere it might depend on the same project that requires `node-config`, but is "version 2" of the same project, and the different versions need different configurations, correct?

This sounds a lot like #225 and #226 Have you read through those related issues?

vifird:
@markstos Yeah, almost the same problem, except a different: the submodules is the current repository itself, but different commits. So the `setModuleDefaults` doesn't fit, for the same commit may be main module, also may be submodule. Besides, the setModuleDefaults need to change `NODE_CONFIG_DIR` and clean the cache, this usage invades codes too much i think.

The usage of `configrc` is like this: 
```
|-- server.js
|-- configrc.json5
    : { NODE_CONFIG_DIR: "./config" }
|-- config
    |-- default.json
    |-- production.json
|-- submodules
    |-- version_1.1.6
        |-- entry.js
        |-- config
            |-- default.json
            |-- production.json
        |-- configrc.json5
            : { NODE_CONFIG_DIR: "./submodules/version_1.1.6/config" }
        |-- node_modules
            |-- node-config
    |-- version_1.1.8
        |-- entry.js
        |-- config
            |-- default.json
            |-- production.json
        |-- configrc.json5
            : { NODE_CONFIG_DIR: "./submodules/version_1.1.8/config" }
        |-- node_modules
            |-- node-config
```
Each submodule has its own configurations directory, own node_modules which contains `node-config`, also own `configrc.*` file which leads `NODE_CONFIG_DIR` to their own configurations directory.
Then run `USE_CONFIG_RC=true CONFIG_RC=configrc.json5 node server` to start.
Remark: the `configrc.*` file can be copied into every submodule by outside respositoy.

markstos:
I'm closing this in favor of #670, which offers a clearer vision of how to address submodule support.

I'm not sure #670 addresses having different versions of configuration for different versions of the same module, but no else but the author has expressed on this PR since 2018, so it is apparently an edge case.

PR #466, opened 2018-02-05
allows both coffeescript v1 and v2 to parse .coffee config files

donnut:
Coffeescript version 2 has its dash removed from the module name. This PR allows both `coffee-script` and `coffeescript` as module name.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -34,6 +34,7 @@ var DEFAULT_CLONE_DEPTH = 20,
 
 // Define soft dependencies so transpilers don't include everything
 var COFFEE_DEP = "coffee-script",
+COFFEE2_DEP = "coffeescript",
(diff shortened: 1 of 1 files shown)


markstos:
I'm OK with the concept of this, but is this ordering we want?  The current patch is backwards compatible-- a plus. But generally I would expect we would want to try the *newer* module first before trying the older one.

If the "version 2" parser will handle parsing existing config files in backwards compatible way, I would prefer trying it first.

donnut:
Checking first for the newer version makes sense too. I'm not sure it will make any difference in most situations as I expect people only include one version in there package.json file. 

Shall I reverse the order?

markstos:
A version of CoffeeScript 2 support was released in May 2018 in 1.31.

Issue #467, opened 2018-02-12
config.get() and nested raw


akaNightmare:
### I'm submitting a ... 
  - [  ] bug report
  - [x] feature request
  - [  ] support request or question => Please do not submit support request or questions here, see note at the top of this template.

### What is the current behavior?
`default.js`
```js
const { raw } = require('config/raw');

module.exports = {
    logger: {
        name: 'test',
        streams: [
            {
                type: 'stream',
                stream: raw(process.stdout),
                level: 'debug',
            },
        ],
    },
};
```
`some-other.js`
```js
const config = require('config');
console.log(config.get('logger'));
// ACTUAL: { name: 'test',
  streams: [ { type: 'stream', stream: [RawConfig], level: 'debug' } ] }
```

### What is the expected behavior?

// EXPECTED: { name: 'test',
  streams: [ { type: 'stream', stream: [__WritableStream__], level: 'debug' } ] }
### Please tell us about your environment:

 -  node-config version: 1.29.4
 -  node-version: 8.9.4

markstos:
@patrickpilch Are you up for helping with this bug in the `raw()` feature you contributed?

The problem comes up when a `raw()` value is not the direct target of a `config.get()` call, but is part of larger object that was requested.

patrickpilch:
@markstos Yes, I will take a look at this as soon as I can. 

@akaNightmare In the meantime, would the following suffice as a solution in your case? You'll treat the whole object as raw, therefore losing some functionality like nested `get()` and immutability, but it could get you moving forward.

```javascript
module.exports = {
    logger: raw({
        name: 'test',
        streams: [
            {
                type: 'stream',
                stream: process.stdout,
                level: 'debug',
            },
        ],
    }),
};
```

akaNightmare:
@patrickpilch Thank you, however `streams` property is dynamic, that means for diff environments `streams` can be different and it is not the option for me. I made some dirty hack for that:
```js
const logger = bunyan.createLogger({
    ...loggerConfig,
    streams: loggerConfig.streams.map(stream => ({
        ...stream,
        stream: stream.stream.resolve(),
    })),
});

```

markstos:
I look forward to these patches. I'm a `bunyan` user also. I happened to dodge the bullet because I only need `stream` set to the default value of `process.stdout`, therefore I don't need to include the default in the config file.

akaNightmare:
@patrickpilch it does not work with nested raw in array:
```js
[...]

PR #474, opened 2018-02-26
Support for nested raw() in javascript configurations

patrickpilch:
Moving the `raw` wrapped object unwrapping into `makeImmutable`
is piggy-backing on the existing recursive object graph traversal
to ensure that regardless of how deep a raw-wrapped object is
within the configuration, it will be unwrapped.

The ordering of `makeImmutable` and `getImpl` was flipped in order
to consistently return an immutable config. Previously, a mutable
configuration was returned on first `get()`, and an immutable one
on subsequent `get()`s.

Fixes lorenwest/node-config#467

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -186,13 +186,6 @@ Config.prototype.get = function(property) {
   if(property === null || property === undefined){
     throw new Error("Calling config.get with null or undefined argument");
   }
-  var t = this,
-      value = getImpl(t, property);
-
-  // Produce an exception if the property doesn't exist
-  if (value === undefined) {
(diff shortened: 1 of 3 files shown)


markstos:
@lorenwest My only concern was the changed ordering of `makeImmutable` and `getImpl`, but I confirmed that we already have test coverage for " No mutation after the first get()".

I'm in favor of merging this.

lorenwest:
This looks good - thanks for the change.

lorenwest:
Published 2/26/2018 at version 1.30.0

PR #480, opened 2018-05-04
Make the logger customizable

Maxithi:
The aim of this PR is to make the logger customizable in a non-major update.

# Objective
The default behaviour of outputting information using `console.error()` can sometimes not be the desired one. Advanced logging like the one done by [debug](https://www.npmjs.com/package/debug) requires the ability for the user to inject his desired logging method.

# Obstacle
Most modules implement this by adding a parameter `options` when requiring the module and passing it on to the constructor.

```JavaScript
var options = {logger: console.log};
var config = require('config')(options);
```

This would however lead to a breaking change forcing to update the module requirement from `var config = require('config')` to `var config = require('config')()`. This would only be acceptable in a major version update and thus, while waiting for this kind of update, a work-around will be suggested.

Other advantages of a future modification of the module exportation include the ability to have different logging methods for different instances of the `config` object. 

# Usage
In order to pass the logger method to the config object before it finished construction, a global variable can be used.
```JavaScript
// We declare the global variable. Here using the debug module.
global.NODE_CONFIG_LOGGER = require('debug')('config');
// Afterwards, we require config which will check for the global `global.NODE_CONFIG_LOGGER ` variable.
//...
```

The previous usage will allow us to have logging information which will be filtered in [debug](https://www.npmjs.com/package/debug)'s case: `config FATAL: NODE_ENV value of 'test' did not match any deployment config file names. +0ms`  

Logging can also be hot-swapped by simply calling the `util.logger(logger)` method.
```JavaScript
// Here we hot-swap the logger method.
var logger = require('debug')('config');
config.util.logger(logger);
//...
``` 

# Future
As explained previously, this work-around should be removed in the next major update in order to favour the `var config = require('config')({logger: someLoggerMethod})` notation by altering the way the module is exported. As time is of the essence, this workaround allows the changes to be published without requiring a major update and rewriting all tests.

# Alternative
Should however the maintainers think that a major update is acceptable in order to allow further configuration parameters to be passed to the constructor in smaller updates, I'd be happy to implement it and rewrite the tests. This PR and proposition is to prevent useless forking and re-publishing of this module which I'd like to use in an upcoming production environment.


markstos:
I'm in favor of this change, although I would simplify it. 

I would remove the `logger` method. If you want to set a new `log` method:

    config.util.log = config.error;

If you want to get the log method:
  
    var logger = config.util.log;

The final decision will be made by @lorenwest.

lorenwest:
I'm in favor of this PR, and @Maxithi what do you think about the simplification suggested by @markstos?

As for a major upgrade, it's been a while since we've discussed it. I like the concept of independent config objects, and would want to be very clear as to the intention. It's a slippery slope between static configs that you can rely on, and dynamic configs that change every time you read them. We had dynamic configs at one point, and rolled it back in favor of immutability.

I have some thoughts on keeping immutability in conjunction with a constructor. The immutability is fixed at construction time, so while you're using that instance the configs don't change. This could be useful for changing configs at runtime, yet keeping the configs constant for, say, the lifecycle of an inbound request.

Maxithi:
I agree with @markstos 's suggestion as long as we have the ability to define the logger before the object construction in order to canalise the logs properly from the beginning.

In regards to @lorenwest 's concerns about immutability, I do fully support *configuration immutability* while having the possibility to have a *mutable object*. 

In our app's entry point, we should be able to require the module without using it in order to load the appropriate configuration file.
```JavaScript
var logger = require('debug')('config');
```
Any errors or missing files would be logged through [debug](https://www.npmjs.com/package/debug) as `config {message} + 0ms`.

Later, in another file (in my case a Database Manager class), we would be able to require the module with the **same configuration** again but with another debugger.
```JavaScript
var logger = require('debug')('config:database');
```
Any logging from actions within this file through [debug](https://www.npmjs.com/package/debug) would look like `config:database {message} + 0ms`.

The advantage of such logger dependency injection is to allow further ease of updates. Should you later add a *silent* mode for the `get()` method which returns `undefined` rather than throwing when missing an configuration property (which I'd by the way find useful for some cases), you would be able to log the missing property as `config:{namespace} {message} +0`.

How would you like to implement this @lorenwest ? As a small temporary work-around update or breaking one?

lorenwest:
As for multiple instances, I envision each instance accepting any number of config overrides including environment variable related. Once instantiated, that config object would be immutable.

To retain backward compatibility we could instantiate one "static" instance to use when invoked using the existing api.

If interested in pursuing, I suggest creating a separate topic.

lorenwest:
Also, if we create a new major version we could drop Node4 support and your pull request wouldn't fail the Travis checks.

Maxithi:
> To retain backward compatibility we could instantiate one "static" instance to use when invoked using the existing api.

How would you implement these two abilities of constructing static and dynamic instances?

lorenwest:
@Maxithi - NodeJS 4 doesn't support ES6 syntax, so the failure is here
```
    let notifier = require('update-notifier')({pkg})
    ^^^
```

As for supporting a constructor and non-constructor, I would export a constructor function vs. an object (as we are today).

If clients call methods directly on the function, they're using the current syntax. Those methods would use a private global config instance (like the global instance we use today). Here's an example client use (compatible style):
```
let config = require('config');
let my_config_val = config.get('my_config_val');
```
And a new style client:
```
let Config = require('config');
let config = new Config({initial:"parameters"});
let my_config_val = config.get('my_config_val');
```
And the node-config implementation would look something like this:
```
// Constructor exported
let Config = module.exports = function(params) {
    this.configs = ...
    ...use params to set pre-config params such as NODE_ENV, etc
    ...read values from fs like now
    ...overwrite configs with supplied params
}
Config.prototype.get = function(key) {
    return this.configs[key]; // simplified for example
}
Config.get = function(key) {
    return staticConfig.get(key);
}
const staticConfig = new Config();
```
If the API is kept the same, that should allow clients to migrate to the new style without breaking their existing clients that use the current style.

markstos:
@ghost Could you resolve conflicts and test failures? At this point you can consider Node 10 to be the minimum supported version.

markstos:
Closing due to lack of interest for nearly 3 years.

PR #484, opened 2018-05-22
Issue 483 support coffeescript

bastbijl:
A proposed solution discussed in https://github.com/lorenwest/node-config/issues/483


bastbijl:
Please restart the build on travis. It errored on the line I wrote in the History.md

markstos:
This looks like exactly the fix we want.

@lorenwest The only failing Travis check is with Node 4, with is EOL'ed. It's failing not because this code is not  Node 4 compatible, but because `npm` module has dropped support for Node 4, and Travis tries to use `npm` as part of the setup.

I recommend we go ahead and drop support for Node 4 as well. AWS has already deprecated support it in AWS Lambda as well.

lorenwest:
I'm ok with dropping Node 4 support. @bastbijl - if you edit the `.travis.yml` file to remove node 4 support and re-submit, this will likely pass the Travis tests.

bastbijl:
For some reason I had to also go back to a none recent version of coffeescript for the tests to pass.

lorenwest:
@bastbijl - It's probably worth getting to the bottom of the problem with CS versions. If we have some old tests laying around that only work in old or buggy versions of CS, those tests need changing.

bastbijl:
@lorenwest - If that is the case I'm sure you are right, but reading the travis logs it is not tests that fail. It more seems npm is using a cached repo that does not contain the last version of coffeescript.

markstos:
@lorenwest What level of version bump is appropriate since we are dropping Node 4 support?

Also `engines` in package.json should be updated to patch `.travis.yml`.  It still expresses that we support Node.js 4 there.

lorenwest:
@markstos - Good question. I think it's appropriate to do a major version bump as soon as we do something that we know to be V4 incompatible. At this point it's removing support because Travis is a PITA, but I'm sure as soon as we stop testing in V4 things will sneak their way in.

I hate to version bump for this, because everyone not on V4 (ie: everyone) requires a change to their package.json. It seems like a major version bump would be accompanied with some value received in exchange for the work required.

Thoughts?

lorenwest:
Maybe the version bump could be accompanied with the feature of allowing multiple environments in NODE_ENV?

markstos:
@lorenwest Sounds good.  People also only really need to update their package.json if they have some reason to upgrade. So by the time people get around to upgrading to the new major version, perhaps they will have found some compelling reason for them at that time.   

We can also be clear in our Changelog that the new major version is backwards compatible with the old one for Node > v4

PR #486, opened 2018-05-23
Add support for load multiple files by values defined in NODE_ENV

leonardovillela:
See #481, @lorenwest.


leonardovillela:
Hello Guys, any problem with this PR? I can help with something?

lorenwest:
Thank you for your patience - I've been pretty busy lately.

test/14-config/bare-metal.json
```diff
@@ -0,0 +1,5 @@
+{
+  "db": {
+    "name": "bare-metal-config-env-provided"
+  }
+}
```
lorenwest: Looks like you've removed the final newline in these files. Could you un-do those changes?

lorenwest:
This looks good - thank you for adding this feature. Only one comment in the code review - if you could undo those newline removals and re-issue the PR, I'll get it merged.

test/14-config/bare-metal.json
leonardovillela: Yes, i will update this

lorenwest:
@leonardovillela - We are having a conversation on #484 regarding including this in a major version bump. Will get this published as soon as we conclude that version bump issue.

KrishnaPG:
Would `default` be still loaded in case multiple env are specified?

markstos:
I noticed the wiki was never updated to reflect this change. Documenting it now.

Issue #504, opened 2018-09-17
[request] Add optional default value to config.get()


javl:
### I'm submitting a ... 
  - [x] feature request

I tried finding a request similar to this one in the issuetracker/wiki to see if there is a reason why it's not in config yet but I couldn't find anything. Sorry if I missed it.
I added this request to my local version of the code, but I'm not sure about updating the tests, plus that I wondered if there was any special reason I might be missing why something similar hasn't been implemented yet.

### Request
When getting the value of a property that doesn't exist, instead of catching the exception or working with ```config.has(...)```, wouldn't it be useful to add an optional argument to ```config.get(...)``` to specify a default value to use when the given property doesn't exist?
Something like:

    thing = config.get('name', 'foobar');

If the config contains a ```name``` property, ```thing``` will be set to its value. If ```name``` does not exist, instead of throwing the ```Configuration property 'name' is not defined``` exception, ```thing``` will be set to ```foobar```.

Locally, I updated the get function to look like this (relevant parts only):
```
Config.prototype.get = function(property, defaultValue=null) {
  (...)
  // If the property doesn't exist...
  if (value === undefined) {
    // ... produce an exception if no default value was specified
    if (defaultValue === null) {
      throw new Error('Configuration property "' + property + '" is not defined');
    }
    // or return the given default value
    else {
      value = defaultValue;
    }
  }
};
```
### What is the current behavior?
You can get a value from your config file using ```config.get('name')``` which thows an exception if the property does not exist. You can catch the exception and assign a default value for the non-existing property, or you can do something like:

    if(config.has('name')){
        thing = config.get('name');
    }else{
        thing = 'foobar';
    } 

### Please tell us about your environment:

 -  node-config version: 2.0.1
 -  node-version: 7.10.1

lorenwest:
Hi @javl - this issue has some history, but I'm going to keep this discussion open for a while because I think it's an important topic.

We haven't added this feature because I believe default configurations should be contained in one place. When your servers aren't acting the way they should, at 2am, you want someone (possibly not you) to easily understand the config values running in the application. If they have to sift through tens or hundreds of thousands of lines of code to find default values, it adds time to their task.

We have broken our own rule in this regard by suggesting sub-modules can provide their defaults in code, and I'm not very happy with that solution.

This is one of those "because you can, doesn't mean you should" features. I understand other config libraries offer this ability, but node-config is fairly opinionated.

[...]

PR #511, opened 2018-10-16
Feature : Add default value to get method

AbderrahimSoubaiElidrissi:
Now you can set the default value of a property 

````
// with localhost here as default value of  Customer.dbConfig.Host if is null or undefined
var dbConfig = config.get('Customer.dbConfig.Host', 'localhost');

````

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -181,11 +181,19 @@ var getImpl= function(object, property) {
  *
  * @method get
  * @param property {string} - The configuration property to get. Can include '.' sub-properties.
+ * @param defaultValue {string} - The default value if configuration property not set.
  * @return value {*} - The property value
  */
-Config.prototype.get = function(property) {
-  if(property === null || property === undefined){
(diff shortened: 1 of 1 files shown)


lorenwest:
This issue has a sorted history. Search for 'default' in issues and you'll see threads and threads of discussion.  Here's one: https://github.com/lorenwest/node-config/issues/504

markstos:
It's surprising how many people submit basically the same pull request without first searching for existing issues and pull requests first.

markstos:
Also see #381, #393 and #462.

lorenwest:
To be fair, defaulting on get() appears to be an obvious omission to the newcomer, and it pains me to see people putting effort into this, just to be rejected by those of us that see it so many times.

I would accept a pull request with a comment near Config.prototype.get() that describes the issue, with a search URL for more info.

markstos:
I agree it could be a helpful hint for the next person who considers adding the feature.

PR #516, opened 2018-12-01
Fixed property mutation. Throw an exception on such an attempt. Updat…

fgheorghe:
…ed tests to capture this exception.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -429,19 +429,28 @@ util.makeImmutable = function(object, property, value) {
         value: Object.freeze(value)
       });
     } else {
+      // Call recursively if an object.
+      if (util.isObject(value)) {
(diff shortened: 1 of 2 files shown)


lorenwest:
This looks great.  I'm not as familiar with Proxy and Reflect as you are, so I can't give it a thorough review, but if it passes the tests and works on all versions tested by Travis it should be good.

lorenwest:
I'll give it a day or two for others to review, then merge.

fgheorghe:
Sounds good - thank you!

markstos:
Apologies for the delayed merged.

It's helpful to explicitly throw errors when mutation attempts are made. This may need a major version bump as a behavior change, though.

markstos:
@fgheorghe Could you check out the linked bug reports related to this change?

Issue #518, opened 2018-12-10
setModuleDefaults incorrectly adds a second nested object of the same module name


exogen:
### I'm submitting a ... 
  - [x] bug report

### What is the current behavior?

The config object resulting from a call to `setModuleDefaults` has the module name nested as a key in the module's own config, set to an empty object. Repro:

```js
const config = require('config');

config.util.setModuleDefaults('TestModule', { foo: 'bar' });

// This should be an error, but returns {}.
config.get('TestModule.TestModule');

console.log(config);
// Config { TestModule: { foo: 'bar', TestModule: {} } }
//                                    ^^^^^^^^^^^^^^
```

I believe the bug is due to this line:

https://github.com/lorenwest/node-config/blob/78f88fbd9855131dea0aea23258abb9f992716cb/lib/config.js#L286

This line doesn't make sense to me, as it's extending `moduleConfig` at the path `moduleName.split('.')` – which means the object we just passed into `setModuleDefaults` is going to get a key with its own name in it, for seemingly no good reason. If anything, it seems like this line should apply to the base config, not `moduleConfig`.

### What is the expected behavior?

The configured module object does not need an extra nested key with its own name. Output above should be:

```js
console.log(config);
// Config { TestModule: { foo: 'bar' } }
```

### Please tell us about your environment:

 -  node-config version: 3.0.0
 -  node-version: 10.13.0

lorenwest:
This certainly does look like a bug introduced in #498. @bertho-zero - would you like to address this issue?

bertho-zero:
I did not change the behavior, I just added the depth support and extended the test.

It seems to me that it comes from here:
https://github.com/lorenwest/node-config/blob/78f88fbd9855131dea0aea23258abb9f992716cb/lib/config.js#L285-L290

exogen:
I confirmed that the issue didn't exist in parent commit before #498, so this was introduced along with that depth support. Luckily the fix appears as simple as removing that line, posted in #520.

PR #520, opened 2018-12-14
Don't extend the input module config with its own key

exogen:
* Remove unnecessary line causing the bug.
* Added a new regression test for this.
* Existing tests still pass.

Fix for #518.


lorenwest:
Without knowing the reason the line was there in the first place, I was hesitant to merge. Then I thought if it was important it should have had an associated test. 

Merged.

lorenwest:
Published today in 3.0.1 - thanks for the bugfix!

Issue #531, opened 2019-01-24
Idea: shell-style < operator to read file contents


leoj3n:
I know the wiki suggests using `.js` and `fs.readFileSync`:

https://github.com/lorenwest/node-config/wiki/Special-features-for-JavaScript-configuration-files#including-other-files

But often I just need to quickly read a key/cert file and found myself wishing I could just do:

```json
{
  "key": "< ../ssl/key.pem",
  "cert": "< ../ssl/cert.pem"
}
```

In my `development.json` for example. This seemed so natural to me I thought I might suggest it here.

Is file resolve already supported in this package? I come from using [`@feathersjs/configuration`](https://github.com/feathersjs/feathers/tree/master/packages/configuration) that does [this](https://github.com/feathersjs/feathers/blob/cd4b64419cea4f6c0a89a026a3c0a8875eee4996/packages/configuration/lib/index.js#L28-L34) so maybe this makes more sense to suggest there.

leoj3n:
@lorenwest Cool, that would save me from having to do https://github.com/lorenwest/node-config/issues/253#issuecomment-457035455 in all my configs.

iMoses:
@markstos can be considered an alternative parser which can be defined with `Parser`.
Feels to me that at this point if we would like to create a template language it's best to be a part of a sibling package and not this one.

leoj3n:
HCL (Hashicorp Markup Lang) might be relevant to this discussion as well?

On July 11, 2019 6:07:27 PM CDT, iMoses <notifications@github.com> wrote:
>@markstos can be considered an alternative parser which can be defined
>with `Parser`.
>Feels to me that at this point if we would like to create a template
>language it's best to be a part of a sibling package and not this one.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/lorenwest/node-config/issues/531#issuecomment-510685555

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

iMoses:
@leoj3n you can simply add HCL support if you have a parser to translate HCL to JSON, or you're willing to create one.

For example, use can use [hcl-to-json](https://www.npmjs.com/package/hcl-to-json)
```
const hcltojson = require('hcltojson');
const Parser = require('config/parser');
Parser.setParser('hcl', (filename, content) => hcltojson(content));
module.exports = Parser;
```

markstos:
Closing in favor of @iMoses suggestion.

[...]

PR #533, opened 2019-01-29
feat: shell-style read

leoj3n:
#531

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -658,6 +658,7 @@ util.loadFileConfigs = function(configDir) {
   util.extendDeep(config, runtimeJson);
 
   util.resolveDeferredConfigs(config);
+  util.shellStyleRead(config);
 
   // Return the configuration object
   return config;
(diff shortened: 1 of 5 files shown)


lib/config.js
```diff
@@ -702,6 +703,53 @@ util.resolveDeferredConfigs = function (config) {
...
 }
 
+// Using basic recursion pattern, find all the shell-style read values and get file contents.
+util.shellStyleRead = function (config) {
+  var completeConfig = config;
+
+
+  function _iterate (prop) {
+
+    // We put the properties we are going to look it in an array to keep the order predictable
```
lorenwest: typo `look it` should be `look at`

lib/config.js
```diff
@@ -702,6 +703,59 @@ util.resolveDeferredConfigs = function (config) {
...
+    // First step is to put the properties of interest in an array
+    for (var property in prop) {
+      if (prop.hasOwnProperty(property) && prop[property] != null) {
+        propsToSort.push(property);
+      }
+    }
+
+    // Second step is to iterate of the elements in a predictable (sorted) order
+    propsToSort.sort().forEach(function (property) {
+      if (prop[property].constructor == Object) {
```
jdmarshall, line 731: What do you get by processing the entries in an order other than what they were defined in?

Seems like it could be hiding information from people trying to debug a failure during config loading

markstos:
I'm closing this in favor of the custom parser support added in 3.2.0:
https://github.com/node-config/node-config/wiki/Modifying-the-Configuration-Files-Parser

If we merged this, there would be backcompat issues, if people had defined a key/value pair like   `headerHtml: "<h1>BIG</h1>"`.

It seems that if you want publish a custom parser that does shell-style reads for others to use with node-config, you could do that.

PR #535, opened 2019-02-04
Added ability to use environment vars for partial string replacement

janez-podhostnik:
Added ability to use *templates* inside **custom-environment-variables.json** (yaml, coffe, ect.).

### example:

```json
{
  "logDestination": "/somewhere/{{NODE_ENV}}/log.txt",
  "app1InternalURL": "http://{{ROOT_URL}}/app1/",
  "app2InternalURL": "http://{{ROOT_URL}}/app2/"
}
```
if **ROOT_URL** is **localhost** and **NODE_ENV** is **test** the config values would be overridden by:

```json
{
  "logDestination": "/somewhere/test/log.txt",
  "app1InternalURL": "http://localhost/app1/",
  "app2InternalURL": "http://localhost/app2/"
}
```

Behavior is the same as with normal replacements, meaning that if any of the environment variables that are required in the string are missing the value will not be overridden. You can have as many replacements in one string as desired.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,30 @@
+3.2.3 / 2019-10-03
+==================
+
+  * Fixed strict mode warning #460 - @fedulovivan
+
+3.2.2 / 2019-07-20
+==================
+
+  * Fixed delimiter bug in configDirs to match O/S delimiter - @iMoses
+
+3.2.1 / 2019-07-18
+==================
+
+  * Fixed TypeError: obj.toString is not a function - @leosuncin
+
+3.2.0 / 2019-07-11
+==================
+
+  * Asynchronous configs - @iMoses
+  * Multiple config directories - @iMoses
+  * Improved parser support - @iMoses
+
+3.1.0 / 2019-04-07
+==================
+
+  * Support of module.exports syntax for TS config files @keenondrums
+
 3.0.1 / 2018-12-16
 ==================
 
--- a/History.md
+++ b/History.md
@@ -1,3 +1,8 @@
+3.0.1 / 2018-12-16
+==================
+
+  * Fixed bug where dot notation extended own key @exogen
+
 3.0.0 / 2018-11-20
 ==================
 
--- a/README.md
+++ b/README.md
@@ -151,19 +151,19 @@ Contributors
 <td><img src=https://avatars2.githubusercontent.com/u/2529835?v=4><a href="https://github.com/simon-scherzinger">simon-scherzinger</a></td>
 <td><img src=https://avatars1.githubusercontent.com/u/175627?v=4><a href="https://github.com/axelhzf">axelhzf</a></td>
 <td><img src=https://avatars3.githubusercontent.com/u/7782055?v=4><a href="https://github.com/benkroeger">benkroeger</a></td>
-<td><img src=https://avatars3.githubusercontent.com/u/1443067?v=4><a href="https://github.com/IvanVergiliev">IvanVergiliev</a></td>
-</tr><tr><td><img src=https://avatars0.githubusercontent.com/u/1736957?v=4><a href="https://github.com/jpwilliams">jpwilliams</a></td>
+<td><img src=https://avatars3.githubusercontent.com/u/1872824?v=4><a href="https://github.com/fgheorghe">fgheorghe</a></td>
+</tr><tr><td><img src=https://avatars3.githubusercontent.com/u/1443067?v=4><a href="https://github.com/IvanVergiliev">IvanVergiliev</a></td>
+<td><img src=https://avatars0.githubusercontent.com/u/1736957?v=4><a href="https://github.com/jpwilliams">jpwilliams</a></td>
 <td><img src=https://avatars2.githubusercontent.com/u/1246875?v=4><a href="https://github.com/jaylynch">jaylynch</a></td>
 <td><img src=https://avatars1.githubusercontent.com/u/145742?v=4><a href="https://github.com/jberrisch">jberrisch</a></td>
 <td><img src=https://avatars1.githubusercontent.com/u/9355665?v=4><a href="https://github.com/kgoerlitz">kgoerlitz</a></td>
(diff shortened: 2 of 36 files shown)


leoj3n:
Was wondering why something like this wasn't supported already. Can't find any discussion directly about this, and can't think of any reason not to do this, as it's basically the same replacing once.

However, what if someone actually wanted to use `{{}}` as a literal in the config value? Probably `${}` would be less likely to conflict since people know that from [JavaScript template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

This affects backwards compatibility in the same way as #533 so you probably need to add a flag.

lorenwest:
I'm a bit surprised there hasn't been more discussion about template variables. 

You've done a good thing by limiting it to environment variables, but as sure as I'm writing this, it will be requested to expand to other template features, expand beyond environment variables.

It's probably not needed because, after all, this is a scripting language. It doesn't go un-noticed that all you have to do is use a .js file rather than .json (etc) and use JS templating or your favorite (opinionated) templating language.

I'd like to be convinced there's a need for this before merging.

leoj3n:
> You've done a good thing by limiting it to environment variables

To clarify for anyone else passing by, this means the template replacement only affects values in:

- `custom-environment-variables.*`

That's unfortunate because, in unison with #533, I could imagine wanting to do something like:

```json
{
  "key": "< ../ssl/${NODE_ENV}/key.pem",
  "cert": "< ../ssl/${NODE_ENV}/cert.pem"
}
```

Would allowing this open can of worms similar to #318? I'm not sure.

leoj3n:
@lorenwest Is there a reason why [env vars](https://github.com/lorenwest/node-config/wiki/Environment-Variables#custom-environment-variables) are only replaced in `custom-environment-variables.*`?

Coming from https://docs.feathersjs.com/api/configuration.html I don't really see a reason other than forced opinionated organization. I get setting up configuration in stages, but I don't understand why utilizing env vars would need to be its own step.

lorenwest:
@leoj3n - You may want to ask the author of this PR @janez-podhostnik 

I also have questions about this PR that need to be addressed before merging.

lorenwest:
Opinions follow. Read at your own risk...

I generally see people in two camps - those that see beauty in the stability of configs, and those that see beauty in the flexibility of configs. When you add the ability to process templates, you're addressing that second crowd.

If you're in the second crowd, you're not opposed to using .js files vs. .json files to store your configurations. Javascript has built-in template replacement using `${variable}` 

With that said, I'm questioning the need for this PR. If you like the flexibility of templates, use .js to store your configurations, and do whatever you want in the .js file including template replacement, string concatenation, function calls and regex's to your heart's content.

This has been the way node-config has worked all along. Strictness and stability is for .json, flexibility is for .js.

I'm not sure I see a need to place flexibility into the format designed for people that don't see beauty in that flexibility, when it's already supported by using .js configuration files.

Does that make sense?

leoj3n:
> Correct me if I'm wrong, but as far as I saw you could still do that by putting
> 
> ```json
> {
>   "key": "< ../ssl/${NODE_ENV}/key.pem",
>   "cert": "< ../ssl/${NODE_ENV}/cert.pem"
> }
> ```
> 
> inside of custom-environment-variables.* I will double check and if this is desired I will make sure the code works that way.

You're right, @janez-podhostnik, using `NODE_ENV` here was a bad example for what I was trying to say, because of support for `NODE_ENV.*` files.

However, @lorenwest, I still think there is a use case for allowing env var substitution in files other than `custom-environment-variables.*` like supported by <https://docs.feathersjs.com/api/configuration.html> for doing things like:

```console
$ cat server/config/development.json
{
  "port": 3030,
  "host": "DEV_SERVER_HOST",
  "mongodb": "mongodb://localhost:27017/my-app"
}

$ cat server/config/production.json
{
  "port": "PORT",
  "host": "my-app.herokuapp.com",
  "mongodb": "MONGODB_URI"
}
```

At that point most projects could just get rid of `custom-environment-variables.*`, but that's basically a separate issue from supporting mid-string env var replacement (which is why I don't think it's a question for @janez-podhostnik).

janez-podhostnik:
@lorenwest I would still like to keep the simplicity of a json or yaml while having a bit more power for customization. .js files would perhaps lead us in the wrong direction, because of the freedom they give. But that is just my opinion.

lorenwest:
I can see your point about .js being too flexible and limiting template replacement to environment variables only. 

Help me understand why you would want this only in `custom-environment-variables.*` and not everywhere. It seems like this feature renders `custom-environment-variables.*` irrelevant.

janez-podhostnik:
I wanted to limit my changes as much as possible and since `custom-environment-variables` are already used only for environment variables it seemed sensible to limit myself to that file.

iMoses:
@janez-podhostnik if you're interested, I just posted a proposal which enables middleware and can be used to create a cross-extensions template engine. 
#569 I'm still unsure about what features should the template engine support (e.g. partials), so if you have an opinion on the subject I'd be more than happy to hear it as part of the discussion.

youyinnn:
any progress here?

lorenwest:
I've been holding out on new discussions since my opinions were stated above. I'm still of the opinion that templates belong in .js files where templates are directly supported. The reasons for not supporting templates in non-.js files is the same reason people have for choosing those formats.

In fact, people may want to put templates in strings in those formats for the specific purpose of having those NOT expanded by node-config, but later in the process under their own control. Having node-config try to replace those will significantly change the way they use node-config.

I'm open for further discussion, but I'm wanting to close this issue if it continues to generate work for the maintainers.

KernelDeimos:
following your point that people may want to process config values their own way, opt-in middlewares to transform config values might be a good approach.

markstos:
I'm closing this, aligning with @lorenwest That you can use JavaScript config files if you want to use a templating system.

The proposal in #569 shows another approach which also keeps templating code out of the code: support parsing middleware which would allow external code to hook into the parsing stage to apply things like middleware. 

Currently, the volunteers maintaining node-config are already stretched thin to support the code that's already in the core. 

As I mentioned in #569, that code could be published and used as a fork.

Issue #539, opened 2019-02-17
Plugin: AWS secrets


leonprou:
I wrote a plugin that taking configurations from AWS Secretes Manager and merging them into the config object: you can see an initial version [here](https://github.com/leonprou/config-aws-secrets). An Example of the usage will follow soon.

I copied some code from the source, so the solution is a little ad-hoc. The best way to generalise it is by a little refactor to separate some functionality to independent methods. Will you be interested in such refactor? I think it's a great project, and it can be improved by user customised plugins.

### I'm submitting a ... 
  - [x] plugin
  - [x] feature request

lorenwest:
We don't really have a plugin mechanism, but that could be a good feature to add.

In the category of merging configs from other locations, the way to do it is to set those values into environment variables, and use the [Custom Environment Variables](https://github.com/lorenwest/node-config/wiki/Environment-Variables#custom-environment-variables) logic already built-in.

We may be upgrading that in the future to allow for `${env_var}` template variables, but the above solution already works without changing any code.

markstos:
With the current functionality, this pattern would be possible:

 1. Create a tiny, private module: `my-config`.
 2. With this module, first load the AWS secrets and set them in environment variables as needed.
 3. Load `node-config`, using the secrets found in the environment variables, via custom-environment-variables.json
 4. Export `config` from `my-config` to the caller
 5. In your project code, refer to `my-config`, which will contain a `node-config` object which includes the AWS secrets as well. 

With this approach, it would also be easy to add a bit of custom code to some something like using the AWS secrets only in production, with static values for local development and testing. 

What is the specific plugin API that's being proposed here?

leonprou:
@markstos @markstos That's interesting I did it differently:

1. Loaded the config.
2. Passed the config object to my plugin
3. the plugin loaded the secrets from AWS and updated the config using:

```
      const environmentSubstitutions = config.util.substituteDeep(configObject, secretVariables)
      config.util.extendDeep(config, environmentSubstitutions)
```

I chose to do it that way reading this wiki [page](https://github.com/lorenwest/node-config/wiki/Configuring-from-an-External-Source). Also, I use the config params to load the AWS secrets, so I needed to require the config module first.

leonprou:
@lorenwest @markstos I did a quick PR to show API changes I want to make: https://github.com/lorenwest/node-config/pull/541.

[...]

PR #541, opened 2019-02-24
Custom env vars refactor

leonprou:
I did some refactors based on this issue: https://github.com/lorenwest/node-config/issues/539.

Did it really quick to get your opinion, so docs and tests are missing. Also tried to keep the changes minimal, just added stuff.


markstos:
Thanks for the contribution but this PR is being closed. See #539 for rationale.

PR #545, opened 2019-03-28
[Closes #524] Feature Added: support of module.exports syntax for TS config files

### LINKED ISSUE #524: Incorrect handling of typescript config files
**Note: for support questions, please use [StackOverflow](https://stackoverflow.com/questions/tagged/node-config) and tag your question with `node-config`**. *This repository's issues are reserved for feature requests and bug reports.*

Before submitting a bug report, please search the issue tracker the wiki first. Many issues have already been discussed.

The wiki is located at: https://github.com/lorenwest/node-config/wiki

### I'm submitting a ... 
  - [ x ] bug report
  - [  ] feature request
  - [  ] support request or question => Please do not submit support request or questions here, see note at the top of this template.

The current implementation for ts config files is incompatible with the handling of js config files, as the ts handling foresees a 'default' named export. This has the consequence that, for a nodejs project, the very same code cannot be made to run both in a typical development setup (where the application is executed with ts-node) and in a production setup where the ts files (including the ts config files) would be compiled to js and the resulting compiled sw is run with node.

I think the correct approach would be to use the typescript 'export =' instead of the 'export default' schema in the ts config files. In this case the generated js would be exactly equivalent to that of hand-written js files.
The config.js implementation should, in my view, be modified at line 787 from
`configObject = require(fullFilename).default;`
to
`configObject = require(fullFilename)`

In this way, there would be perfect alignment between typescript config files compiled on the fly within config.js, typescript config files compiled offline and read by config.js as plain js config files, and hand-written js config files

 -  node-config version: 3.0.1
 -  node-version: 10.12.0
 
### Other information
I tested the proposed setup in my environment by changing line 787 as indicated above.
exampe of a typescript config file I used:

`
import { IConfig } from './config.interface';

const DEV_CONFIG: Partial<IConfig> = {
    server: {
        host: 'aHost',
        port: 3000,
        rootPath: 'alfa/beta/gamma'
    }
};

export = DEV_CONFIG;
`

This here below is the generated js from offline tsc compilation:

`
"use strict";
const DEV_CONFIG = {
    server: {
        host: 'aHost',
        port: 3000,
        rootPath: 'alfa/beta/gamma'
    }
};
module.exports = DEV_CONFIG;

`

As shown, this way the generated js is exactly the same as one would use if the js file had been hand-written according to the documentation of config.js, therefore there is no need to have typescript config files handled differently with the default export.

keenondrums:


keenondrums:
@lorenwest any chance you could take a look? Or who would be your TypeScript guy to review this one?

lorenwest:
I was hoping someone would step up with some TS expertise to weigh in on the choices you've made in this PR.

From my view this PR doesn't appear to have any negative impacts, so I'm going to take your advise and merge it.

Yes, that makes you my typescript guy :)

lorenwest:
Published in 3.1.0

keenondrums:
@lorenwest happy to be of service, sir!
Thank you! :)

PR #555, opened 2019-06-26
Fixed double-defer bug #266

iMoses:
@markstos This is my suggested solution for fixing #266 aka the double-defer bug **using getters**.
~~Notice that it does add a dependency to support synchronous-promises which prevent the library becoming accessible before all deferred configs has been resolved.~~

Let me know what you think..

### DIFF
--- a/defer.js
+++ b/defer.js
@@ -1,3 +1,5 @@
+var SynchronousPromise = require('synchronous-promise').SynchronousPromise;
+
 // Create a deferredConfig prototype so that we can check for it when reviewing the configs later.
 function DeferredConfig () {
 }
@@ -6,7 +8,14 @@ DeferredConfig.prototype.resolve = function (config, original) {};
 // Accept a function that we'll use to resolve this value later and return a 'deferred' configuration value to resolve it later.
 function deferConfig (func) {
   var obj = Object.create(DeferredConfig.prototype);
-  obj.resolve = func;
+  obj.resolve = function(config, original) {
+    var resolve, promise = new SynchronousPromise(function(res) { resolve = res; });
+    promise.exec = function() {
+      resolve(func.call(config, config, original));
+      return promise;
+    };
+    return promise;
+  };
   return obj;
 }
 
--- a/defer.js
+++ b/defer.js
@@ -5,7 +5,7 @@ DeferredConfig.prototype.resolve = function() {};
 // Accept a function that we'll use to resolve this value later and return a 'deferred' configuration value to resolve it later.
 function deferConfig(func) {
   var obj = Object.create(DeferredConfig.prototype);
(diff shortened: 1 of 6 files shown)


lorenwest:
Not a fan of requiring `synchronous-promise` as a dependency. There will be strong opinions on that library, and making it a direct dependency transfers those opinions onto the `config` module itself.

Many people use `config` in production environments, and when the author of `synchronous-promise` recommends against using it in production environments, we can't say it's ok.

lorenwest:
A different approach using promise chaining vs. depending on first-come-first-served sounds more reliable.

iMoses:
@markstos @lorenwest I haven't tested it much and I need to get off the computer now, but it seems to be passing initial tests to I pushed the code if you want to review it. 
Nothing external. I took the "trick" used to resolve and implemented it in a very simple manner.

markstos:
@iMoses This looks very promising, but I think could use some more test coverage to test a config with a default and override. 

As I understand if you one defer function references another value, then the other value will be resolved right away. In a case where there's a default and an override, can this resolution happen to soon?

Consider where key config keys `a` is defined in a `default.js` config file.  `default.js` also defines `b`, which references `a` by a defer function function. 

In `override.js`, a config key `c` references `b` by a defer function. `b` is not defined in override.js, but would be populated when the full configuration as been defined through cascading overrides. 

I think this *may* already work in your code, because the defer functions don't actually get executed until the configuration has been built, when it is safe to resolve them on-demand.

Overall, looks like an elegant solution.

iMoses:
> As I understand if you one defer function references another value, then the other value will be resolved right away. In a case where there's a default and an override, can this resolution happen to soon?

@markstos notice that we split the logic into two stages.
1. Prepare all DeferredConfig objects to be resolved
2. Resolve all deffered functions after the entire tree has been iterated

This should prevent any such case of race conditioning from happening.

Anyhow I'll improve the tests to handle this cases..

markstos:
I'm appreciative! I opened this bug almost four years ago!

iMoses:
BTW, once this issue is fixed we can improve the `asyncConfig` method with the same logic to enable accessing async values from within an async function and asynchronously wait for their results before you resolve. 

This is an awesome feature. Think about this use case, you're calling for services information and secrets from the DB, and once it arrives you trigger multiple requests to bring each more configuration from each service. All dependencies can be simply resolved from within the config file using nothing but simple functions.

```
async function dbConfig() {
  // services information and required secrets
}
async function userToken(cfg) {
  const { username, secretKey } = await cfg.dbConfig;
  return await fetchToken(username, secretKey);
}
module.exports = {
  dbConfig: asyncConfig(dbConfig),
  userToken: asyncConfig(userToken),
};
```

PR #556, opened 2019-06-27
Fixed #449 - support native promises by default, no need to wrap them in raw

### LINKED ISSUE #449: Add native Promise support
This thread is for discussion around adding native `Promise` support to `node-config`. The feature will be complete and merged when tests added by @scottmas in #447 pass, and there is thorough documentation on both native Promise usage as well as async/await usage.

If you're interested in adding this core functionality, please add a comment to this thread and we will discuss the issues here.

iMoses:
Very little changes in code, validate a promise using `util.isObject(obj) && obj.toString() === '[object Promise]'`.

What do you think?

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1392,7 +1392,7 @@ util.extendDeep = function(mergeInto) {
       }
 
       // Copy recursively if the mergeFrom element is an object (or array or fn)
-      else if (mergeFrom[prop] && typeof mergeFrom[prop] === 'object') {
+      else if (mergeFrom[prop] && typeof mergeFrom[prop] === 'object' && !util.isPromise(mergeFrom[prop])) {
         mergeInto[prop] = util.cloneDeep(mergeFrom[prop], depth -1);
       }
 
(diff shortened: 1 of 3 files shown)


markstos:
Which promise libraries did you test with? 

I mostly use native and bluebird. 

What causes a promise to be stringified as '[object Promise]' ?

iMoses:
> Which promise libraries did you test with?
> 
> I mostly use native and bluebird.
> 
> What causes a promise to be stringified as '[object Promise]' ?

I don't think it's in the spec, at least I couldn't find any reference for it, but NodeJS (V8, so also Chrome) internal implementation makes sure of it.
I haven't tested bluebird, only native promises. Should I add tests for bluebird as well?

markstos:
I did a quick test with bluebird and it stringifies that way as well:

```
const bluebird = require('bluebird');
const foo = new bluebird(() => {}, () => {});
console.log (foo.toString());
````

Maybe we don't need to add another dependency just for that test.

iMoses:
I tend to agree.
If a library such a bluebird claims to provide a polyfill or a replacement for promises, it's their responsibility to make sure it is identical in every way to native promises.
As long as we know bluebird doesn't break this logic, I think we're good to go.

iMoses:
@lorenwest what do you say? can this be merged?

lorenwest:
Agreed.  This is new functionality, so we don't have to worry about backward compatibility. Going forward, if they want to use this functionality, they'll have to use native promises, bluebird, or a promise library that stringifes the same way. Not a huge expectation.

Issue #181, opened 2014-12-02
wish: make config file format support pluggable.


markstos:
Supporting additional file formats should be easy. 

The first step for this task is to decide on the design to use. 
## Possible designs
- [node-rechoir](https://github.com/tkellen/node-rechoir) handles this problem and could be added as dependency, effectively outsourcing the problem. This is what `gulp` uses. `node-config` currently has no dependencies, so this would be a change in this regard. 
- _TODO: Add more possibilities here_

vicary:
Came here when looking for the extension priority.

If you are suggesting users to not create more than one file ext. under the same name, I suggest putting this in the [document](https://github.com/lorenwest/node-config/wiki/Configuration-Files#file-formats).

tamtakoe:
config-parser module may have the following structure:
```
config-parser
|- index.js
|- plugins
   |- json
   |- yaml
   |- cson
   |- ...
```

```js
var configParser = require('config-parser');

//Register custom plugin
configParser.registerPlugin(['toml', 'tml'], function(input) { ... return output });

var config = configParser.parse(file);
```

alexindigo:
@markstos I gave it a try in a separate library (to allow myself to start with a clean slate and try different approaches), and I came up with something that should be powerful and flexible enough for different use cases.

This how it looks:

```javascript
var configly = require('configly');

configly.PARSERS = {
  ini       : ini.parse,
  // have it as a wrapper to prevent extra arguments leaking 
  cson      : function(str) { return cson.parse(str); },
  yml       : function(str) { return yaml.safeLoad(str); },
  // same options as used within `config` module 
  properties: function(str) { return properties.parse(str, {namespaces: true, variables: true, sections: true}); },
  // use json5 instead of `JSON.parse` 
  json      : json5.parse
  // keep the original one 
  js        : configly.PARSERS.js,
};
 
var configObj = configly();
```

It is still a singleton, but the module exports load function,
instead of single config object, so it allows more flexibility,
for the price of extra function call.

ltvan:
I see the suggestion of tamtakoe looks good, simple and easy. Why don't just go with it?

Currently I have an issue that I want to have a custom type with js-yaml, but since you have hard-coded the .load call so I cannot inject my custom type. I tried to wrap it with global.Yaml but you declared `var Yaml` at the top, then the global variable can not be reached.

markstos:
[...]

Issue #500, opened 2018-08-31
How to pass a different yaml parser


Kikobeats:
### I'm submitting a ... 

  - [x] feature request

### What is the current behavior?

Currently, the library is forcing to use `js-yaml` or `yaml` dependency as parser, but you can't a different dependency, for example, [yaml-parser](https://bundlephobia.com/result?p=yaml-parser), that in my opinion is better and smaller.

Also, the way for loading these dependencies is making the code so fragile. 

https://github.com/lorenwest/node-config/blob/master/lib/config.js#L882

These libraries are not part of the dependencies projects, so it doesn't make sense force the user choose between them.

I suggest a simple, [open](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle) and more clean way: just expose an option to be possible specify the yaml parser that you want to use:

```js
{
	yamlParser: require('yaml-parser').safeLoad
} 
```

lorenwest:
The parser used must be defined _before_ files are parsed so they can't be specified in config files.

I suggest creating environment variables that define parser libraries to use for different file types. Something like this:

`export NODE_CONFIG_PARSER_YAML=yaml-parser`

These environment variables could override the defaults, and even let you define parsers for your own file format

`export NODE_CONFIG_PARSER_FUNKY=funky-parser`

That would look for `{filename}.funky` files, and parse them. These parsers would have to have a single method `parse(text)` that returns a parsed JS object from the text.

lorenwest:
This would be a nice addition to the config library - fitting for issue #500. Any takers?

Kikobeats:
@lorenwest notes that your solution only works if the parser loaded is the main function exported.

`yaml-parser.safeLoad` doesn't fit into a environment variable, or you need to provide another environment variable for the method.

lorenwest:
The environment variable could reference a module in your local app that wraps the parser for custom loading, and would export a `parse(text)` method that forwards the call with custom arguments, etc.

export NODE_CONFIG_PARSER_FUNKY=./lib/funkyParserWrapper

markstos:
Does it really matter if the YAML parser is optimal? It's used once when the first second of the app starting up, and then not again after that. 

If you are concerned about the small extra RAM taking up by the YAML parser after it is used, it can be [unloaded from memory](https://stackoverflow.com/questions/6676612/unloading-code-modules).

[...]

Issue #348, opened 2016-09-17
Support HOCON for configuration. (PR welcome)


20zinnm:
It's be awesome if we could use HOCON for our configuration files.

Thanks!

lorenwest:
Anyone interested in adding this support has a good pattern to use for testing and lazy loading.

markstos:
[HOCON](https://github.com/typesafehub/config/blob/master/HOCON.md) sounds similiar to [JSON5](https://www.npmjs.com/package/json5) which we already support.

20zinnm:
HOCON is actually extremely dynamic, more than JSON5. It shifts the burden from the person to the machine; granted, that makes it more complex to parse. I do know libraries and such exist, like https://github.com/typesafehub/config <https://github.com/typesafehub/config>, but I’m not aware of a native JavaScript implementation. It’s a JSON superset so it can’t be too hard.

iMoses:
This is can be solved by addressing #181

markstos:
Closing this in favor of #181.

PR #557, opened 2019-06-27
#181 - Support custom Parsers definitions from an external source

iMoses:
This is my initial proposal. 
I just wanna make sure you're okay with this change before moving forward and adding a new env-var which will be loaded and merged with `util.parsers` before loading configuration files.

The only change to logic offered here is the addition of `util.setParser` and `util.removeParser`. 
Everything else should continue working in exactly the same ways as before.

Note: There's a typo in the commit message. Refers to #181

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -4,19 +4,10 @@
 // http://lorenwest.github.com/node-config
 
 // Dependencies
-var Yaml = null,    // External libraries are lazy-loaded
-    VisionmediaYaml = null,  // only if these file types exist.
-    Coffee = null,
-    Iced = null,
-    CSON = null,
-    PPARSER = null,
-    JSON5 = null,
-    TOML = null,
-    HJSON = null,
-    XML = null,
-    deferConfig = require('../defer').deferConfig,
+var deferConfig = require('../defer').deferConfig,
     DeferredConfig = require('../defer').DeferredConfig,
     RawConfig = require('../raw').RawConfig,
+    Parser = require('../parser'),
     Utils = require('util'),
     Path = require('path'),
     FileSystem = require('fs');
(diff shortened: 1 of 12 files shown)


lorenwest:
I enthusiastically support this refactor. Thank you for the effort.

markstos:
@iMoses Can you provide an example of what the ENV structure would look like?

iMoses:
> @iMoses Can you provide an example of what the ENV structure would look like?

@markstos I'll push a code proposal soon, I'm still thinking about the implementation a bit, would appreciate any input you may have.

@lorenwest I wanted to ask if you're okay with separating the parsing logic to a separate file.

I thought that we can add an env-var like `NODE_CONFIG_PARSER_PATH` which will default to `config/parser` but could be changed into a different file. If that's the case, users could import `config/parser` into their custom parser file to get relevant utils and our predefined parsers,

```
import Parser from 'config/parser';
Parser.setParser('jsx', Parser.jsParser);
Parser.setParser('custom', function(filename, content) { /* ... */ });

Parser.getOrder();  // ['js', 'json', 'yaml', 'yml', 'jsx', 'custom']
Parser.getOrder('yaml');  // 2
Parser.setOrder('jsx', 2);  // change file type index in order
Parser.getOrder();  // ['js', 'json', 'jsx', 'yaml', 'yml', 'custom']
```

It would cause a minor breaking change inside the `util` object, which we can cover up for now by linking the references, to keep backwards compatibility.

Also, any input about the API?

lorenwest:
Separating `config/parser` is a good approach, and linking references in `util` to keep backward compatibility is better than the other option of bumping up the major revision.

lorenwest:
The API for get looks reasonable. Mutating the list would be a bit simpler if you replaced the entire array vs. exposing methods for insert/replace/delete. A potentially common pattern would be to lock down parsers to a set number, and using custom insert/replace/delete methods is painful vs. one replace method that replaces the entire array.

iMoses:
What about supporting both?
I updated the code above..

iMoses:
@lorenwest I pushed a new version which implements a full solution + basic tests

What different kinds of tests would you like to see for such a feature?
Once we settle on all the details I'll add new docs to communicate the feature.

iMoses:
Examples of solving relevant issues:

#483
```
import Parser from 'config/parser';
Parser.setParser('coffee', filename => {
  require('coffeescript/register');
  return require(filename);
});
export default Parser;
```

#500
```
import Parser from 'config/parser';
function yamlParser(filename, content) {
  return require('yaml-parser').safeLoad(content);
}
Parser.setParser('yml', yamlParser);
Parser.setParser('yaml', yamlParser);
export default Parser;
```

#348
```
import Parser from 'config/parser';
Parser.setParser('hocon', (filename, content) => require('hocon-parser')(content));
export default Parser;
```

markstos:
@lorenwest @iMoses Lots of good updates. Merged. Is it about time to update History.md and release?

PR #559, opened 2019-06-28
Improved asyncConfig to support new deferConfig mechanism

iMoses:
Now every async function used by `asyncConfig` is also deferred, which means you can access the resolved config including other asyncConfig promises (or native ones), and dependencies are automatically resolved.

Example:
```
async function dbConfig() {
  // services information and required secrets
}
async function userToken(cfg) {
  const { username, secretKey } = await cfg.dbConfig;
  return await fetchToken(username, secretKey);
}
module.exports = {
  dbConfig: asyncConfig(dbConfig),
  userToken: asyncConfig(userToken),
};
```

### DIFF
--- a/async.js
+++ b/async.js
@@ -1,22 +1,35 @@
+var asyncSymbol = Symbol('asyncSymbol');
 var deferConfig = require('./defer').deferConfig;
-
-function AsyncConfig () {}
-AsyncConfig.prototype.getPromise = function () { return Promise.resolve(); };
 
 /**
  * @param promiseOrFunc   the promise will determine a property's value once resolved
  *                        can also be a function to defer which resolves to a promise
- * @returns {AsyncConfig}   a 'async' configuration value to resolve later with `resolveAsyncConfigs`
+ * @returns {Promise}     a marked promise to be resolve later using `resolveAsyncConfigs`
  */
-function asyncConfig (promiseOrFunc) {
+function asyncConfig(promiseOrFunc) {
   if (typeof promiseOrFunc === 'function') {  // also acts as deferConfig
     return deferConfig(function (config, original) {
-      return asyncConfig(promiseOrFunc.call(config, config, original));
+      var release;
+      function registerRelease(resolve) { release = resolve; }
+      function callFunc() { return promiseOrFunc.call(config, config, original); }
+      var promise = asyncConfig(new Promise(registerRelease).then(callFunc));
+      promise.release = release;
+      return promise;
     });
   }
-  var obj = Object.create(AsyncConfig.prototype);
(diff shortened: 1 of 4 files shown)


markstos:
The described functionality is what's desired. I believe when configs are designed to reference other configs, the intent is always reference the *final* (deferred) value of the other configuration.

markstos:
So,  `config` will quickly load synchrously, then when `get()` is called for the first time on an async value, the async/promise code will run, and but follow-up calls to `.get()` will return cached values quickly?

And user's always need to use `await` on these async config values, whether it's the first call or a future call, correct?

iMoses:
@markstos You are correct, but let me layout the processes anyhow, for clarity.

`deferConfig` follows these steps:
1. In `resolveDeferredConfigs` we replace all `DeferredConfig` instances with a getter that's lazily resolves by calling the deferred function.
2. We execute the deferred functions and automatically replace the getter with the function's return value, so the getter will only be called once.

`asyncConfig` is a bit more complicated as it supports both async functions and promises.

In case of an async function we follow these steps:
1. We create a new promise with an external `release` method to resolve it.
    Only on release the async functions are executed, as part of the promise chain.
2. We pass that promise back to `asyncConfig` and return the result.  (see handling promises)
3. We return a `deferConfig` which wraps this logic and resolves to `asyncConfig` value. 

In case of a promise, or an async function that has been wrapped in a promise, we follow these steps:
1. Mark it as asyncConfig using a symbol, to detect later on `resolveAsyncConfigs`.
2. Register a `prepare` method who releases the promise and returns a function which resolves it.
    Resolving means replacing the promise with it's returned value.
3. Return the modified promise.

Then on `resolveAsyncConfigs` we:
1. Collect all marked promises and `prepare` them, which mean executing the async functions.
2. Once all promises been prepared and resolved, we loop over and execute the `resolvers` who replace the promises with their corresponding values.

iMoses:
Technically. if you never use `resolveAsyncConfigs` then you can still use `asyncConfig`s as native promises., though this should come with a warning because using `resolveAsyncConfigs` after we used `get()` would probably be a bug since the config will be frozen.

iMoses:
@lorenwest unrelated question, why is `json5` in `dependencies` and not in `devDependencies`?

lorenwest:
@iMoses - the original `config` library had a filter that removed comments from JSON files prior to parsing, as comments are desirable in config files.

Somewhere around the time we added the JSON5 parser, we decided it was better to parse config JSON files using JSON5 than to support a non-standard JSON format and custom parser.

iMoses:
@lorenwest I think we're good to go, this can be merged and you can feel free to release a new stable version whenever

PR #561, opened 2019-06-29
#560 Improve performance + Support multiple config directories + Fix #512

### LINKED ISSUE #512: Files loaded using NODE_APP_INSTANCE must have matching extensions
### I'm submitting a ... 
  - [ X ] bug report

Using `NODE_APP_INSTANCE` seems to only allow files with the same extension as the equivalent stage config file.  For example, if you have `development.json` and `development-{NODE_APP_INSTANCE}.js`, `config` will not load the values from `development-{NODE_APP_INSTANCE}.js`.  

Reproduction steps:
1. Create a `development.json` file, with some values.  
1. Create a `development-foo.js` file, and export an object that overrides the values defined in `development.json`
1.  `require('config')` in application, setting `NODE_APP_INSTANCE=foo` as environment variable

Actual Behavior: 
Value in `config` will be what was defined in `development.json`

Value in `config` should be what was defined in `development-foo.js`

 -  node-config version: 2.0.1
 -  node-version: 8.11.2

iMoses:
Addresses #560

1. Improved logic to first read directory files and filter those which are relevant to us, then we only try to read them.
2. Allow passing `CONFIG_DIR` with multiple directories separated by `:`. (like PATH)
3. Make sure `NODE_APP_INSTANCE` is loaded on all supported extensions.

Notice that current resolution order won't be separated to each directory but according to files resolution rules. The directory is only used to sort between exact files in separate directories, then the directory index in `CONFIG_DIR` is used to sort.
```
util.locateMatchingFiles('a:b', {'default': 1, 'local': 2})
a/default
b/default
a/local
b/local
```
What do you think  the logic should be in case we add support for multiple directories?

**Edit** I believe the ideal use case would look like that:
```
config/default.js
config/production.json
config/production-instance.json
srv-config/hostname.yaml
srv-config/hostname-instance.yaml
srv-config/hostname-production.yaml
srv-config/hostname-production-instance.yaml
local-config/local.json5
local-config/local-instance.js
```
OR
```
config/default.js
config/production.json
electron-local-config/hostname.yaml
electron-local-config/hostname-production.yaml
config/local.json
```
This feature gives a lot of control to DevOps, and personally I'd be happy to have it supported by this library. It would provide me with much more flexibility to defined my application config distribution.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -566,7 +566,7 @@ util.loadFileConfigs = function(configDir) {
   var baseNames = ['default'].concat(NODE_ENV);
 
   // #236: Also add full hostname when they are different.
-  if ( hostName ) {
+  if (hostName) {
     var firstDomain = hostName.split('.')[0];
 
     NODE_ENV.forEach(function(env) {
@@ -574,7 +574,7 @@ util.loadFileConfigs = function(configDir) {
       baseNames.push(firstDomain, firstDomain + '-' + env);
 
       // Add full hostname when it is not the same
-      if ( hostName != firstDomain ) {
+      if (hostName !== firstDomain) {
         baseNames.push(hostName, hostName + '-' + env);
       }
     });
@@ -584,27 +584,24 @@ util.loadFileConfigs = function(configDir) {
     baseNames.push('local', 'local-' + env);
   });
 
-
+  var allowedFiles = {};
+  var resolutionIndex = 1;
   var extNames = Parser.getFilesOrder();
   baseNames.forEach(function(baseName) {
     extNames.forEach(function(extName) {
-
-      // Try merging the config object into this object
-      var fullFilename = Path.join(CONFIG_DIR , baseName + '.' + extName);
-      var configObj = util.parseFile(fullFilename);
-      if (configObj) {
-        util.extendDeep(config, configObj);
-      }
-
-      // See if the application instance file is available
+      allowedFiles[baseName + '.' + extName] = resolutionIndex++;
       if (APP_INSTANCE) {
-        fullFilename = Path.join(CONFIG_DIR, baseName + '-' + APP_INSTANCE + '.' + extName);
-        configObj = util.parseFile(fullFilename);
-        if (configObj) {
-          util.extendDeep(config, configObj);
-        }
+        allowedFiles[baseName + '-' + APP_INSTANCE + '.' + extName] = resolutionIndex++;
       }
     });
+  });
+
+  var locatedFiles = util.locateMatchingFiles(CONFIG_DIR, allowedFiles);
+  locatedFiles.forEach(function(fullFilename) {
+    var configObj = util.parseFile(fullFilename);
+    if (configObj) {
+      util.extendDeep(config, configObj);
+    }
   });
 
   // Override configurations from the $NODE_CONFIG environment variable
@@ -655,6 +652,25 @@ util.loadFileConfigs = function(configDir) {
 
   // Return the configuration object
   return config;
+};
+
+util.locateMatchingFiles = function(configDirs, allowedFiles) {
+  return configDirs.split(':')
+    .reduce(function(files, configDir) {
+      if (configDir) {
+        try {
+          FileSystem.readdirSync(configDir).forEach(function(file) {
+            if (allowedFiles[file]) {
+              files.push([allowedFiles[file], Path.join(configDir, file)]);
+            }
+          });
+        }
+        catch(e) {}
+        return files;
+      }
+    }, [])
+    .sort(function(a, b) { return a[0] - b[0]; })
+    .map(function(file) { return file[1]; });
 };
 
 // Using basic recursion pattern, find all the deferred values and resolve them.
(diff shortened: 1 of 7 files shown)


iMoses:
I accidentally pushed with the async changes although they're already approved so I'm not sure if it matters or not :)
Let me know if you need me to clean this PR.

iMoses:
I did some thinking and I believe this resolution order _is_ optimal.
It allows me to separate different levels of configurations into different directories and still keep the original resolution order.
e.g. A predefined path of `hostname` configurations on the machine, a predefined location where I put `local` files for specific configurations and the default in-project directory for all project defaults and environment configurations. This is ideal IMO.

My specific use case at the moment is an Electron application which should be able to load overriding configurations from the user's home directory `.config` directory. With the new multi-directory feature that's an easy task. I can have `defaults` and `env` configs as part of my application and override them with `hostname` configs located at `~/.config`.

lib/config.js
```diff
@@ -756,6 +756,9 @@ util.parseFile = function(fullFilename) {
     fileContent = fileContent.replace(/^\uFEFF/, '');
   }
   catch (e2) {
+    if (e2.code !== 'ENOENT') {
+      throw new Error('Config file ' + fullFilename + ' cannot be read');
+    }
```
markstos, line 772: I want to be sure I understand why this change is correct. Before parseFile is called, we will already making a "readdir" call and found which files exist. So, if we are parsing a file and it doesn't exist a moment after it did exist, that's unexpected so we know throw, right?

This a breaking change since we used to return `null` it would be good to document in the exception in the JSDoc comment above the function. 

Also, the line below "return null; // file doesn't exist", but that comment would no longer be accurate, because the "does not exist" case should be covered by ENOENT above.

markstos:
I'm OK with this, although I don't currently use multiple config directories. I guess a parallel would a project like `systemd` that puts "vendor/package" files in `/usr/lib/systemd`, while "local" override files go in `/etc/systemd`.

lib/config.js
iMoses: No, this is to count for cases in which `parseFile` is called from somewhere else, which is not `loadFileConfigs`, since these still happen. This is so we won't make a breaking change :)

Thing is, we got rid of `statsSync` so now we wanna throw an exception only if it's anything other than "file doesn't exist", because we don't care about that. That's the official way suggested by NodeJS.

iMoses: > Also, the line below "return null; // file doesn't exist", but that comment would no longer be accurate, because the "does not exist" case should be covered by ENOENT above.

Notice we check if it is **not** `ENOENT`, then we throw an exception.
We simply ignore cases in which the file doesn't exist.

iMoses:
@markstos I can really use such a feature in two of my current projects.. was actually thinking of developing something on my own, but I'd much rather have it be part of this library.

markstos:
@iMoses Thanks for the clarifications. I endorse the proposal.

markstos:
@iMoses I want to discuss a different matter with. Please provide email or other direct contact to mark@stosberg.com. Thanks.

iMoses:
I'll have a lot of new documentations to write :smile:

iMoses:
@lorenwest you have any input on the changes?

lorenwest:
I support the performance changes, and allowing multiple config directories as segments on `NODE_CONFIG_DIR` seems like a good improvement.

My only concern is that people turn `NODE_CONFIG_DIR` into a `JAVA_PATH` abomination.  I'd like to address the underlying reason for multiple config directories first. If it's a tactical issue - like it allows `config` to be used in electron apps - then it's a good improvement. 

But if the underlying issue is deeper - ie: separate concerns needing to own their own section of the config object - then we may want to go the submodule approach vs. the multi-directory approach.

lorenwest:
Any thoughts on this? I like the simplicity of multiple config dirs with `PATH` syntax, but it doesn't grow as complexity grows. 

Is it a different enough concern from module based configs to call it a feature of its own, or should we use a common module based approach for distributed configurations?

markstos:
I see "multiple directories" as different than module-based configs.  Different use-cases I see:

  * **multiple paths**: A project distributes a project a as `.deb` package with default configuration in `/usr/lib`. The user is expected to override the default configuration with files in  `/etc/`.
  * **module configuration**: A NPM project uses `node-config` internally and uses module configuration for a private configuration. 
  * A complex project uses pulls together multiple private node modules managed  by different teams each with their own separate configuration. I guess either or both approaches might be used in this case.

lorenwest:
Agreed - the `.deb` example is the same as `docker`, and having exactly 2 config directories in the _sealed distribution_ case is very useful.

lorenwest:
Thank you @iMoses for this contribution - it's a significant improvement.

iMoses:
@lorenwest If I may suggest, let's bump the version (I think all these changes deserve a `3.2.0`) before we merge anymore changes :)

ricardolpd:
Apologies for chipping in, i was looking to change the default config path, i found the environment variable, and i end up here. Is there a reason why cant we initialise the config with some options and custom path? Is it an architectural decision, is it because of performance or just choice?

lorenwest:
There are many options and environment variables including changing config path.

Let's move this discussion to [StackOverflow](https://stackoverflow.com/questions/tagged/node-config) and tag your question with `node-config`**. *This repository's issues are reserved for feature requests and bug reports.*

Issue #564, opened 2019-07-12
error for 3.2.0


dingyuanwu:
/admin/xcbasecamp/node_modules/_config@3.2.0@config/lib/config.js:1287
  return util.isObject(obj) && obj.toString() === '[object Promise]';
                                   ^

TypeError: obj.toString is not a function
    at Config.util.isPromise (/admin/xcbasecamp/node_modules/_config@3.2.0@config/lib/config.js:1287:36)
    at _clone (/admin/xcbasecamp/node_modules/_config@3.2.0@config/lib/config.js:975:23)
    at _clone (/admin/xcbasecamp/node_modules/_config@3.2.0@config/lib/config.js:978:20)
    at _clone (/admin/xcbasecamp/node_modules/_config@3.2.0@config/lib/config.js:978:20)
    at _clone (/admin/xcbasecamp/node_modules/_config@3.2.0@config/lib/config.js:978:20)
    at Config.cloneDeep (/admin/xcbasecamp/node_modules/_config@3.2.0@config/lib/config.js:985:10)
    at /admin/xcbasecamp/node_modules/_config@3.2.0@config/lib/config.js:1247:32
    at Array.forEach (<anonymous>)
    at Config.util.extendDeep (/admin/xcbasecamp/node_modules/_config@3.2.0@config/lib/config.js:1222:9)
    at /admin/xcbasecamp/node_modules/_config@3.2.0@config/lib/config.js:603:12

sudosoul:
Uhh.. can you edit your post and include steps to replicate, please?

leosuncin:
I got the same error

**My environmet**
node v11.15.0
config v3.2.0
my config files are TOML

<details>
<summary>Stacktrace</summary>
<pre>
TypeError: obj.toString is not a function                                                                                                                              
    at Config.util.isPromise (/home/user/projects/task-api/node_modules/config/lib/config.js:1287:36)                                               
    at /home/user/projects/task-api/node_modules/config/lib/config.js:1242:21
    at Array.forEach (<anonymous>)                                                 
    at Config.util.extendDeep (/home/user/projects/task-api/node_modules/config/lib/config.js:1222:9)
    at /home/user/projects/task-api/node_modules/config/lib/config.js:603:12                                                                        
    at Array.forEach (<anonymous>)                                                                                                                                     
    at Config.util.loadFileConfigs (/home/user/projects/task-api/node_modules/config/lib/config.js:600:16)                                          
[...]

PR #565, opened 2019-07-12
Fix `TypeError: obj.toString is not a function` in lib/config.js:1287:36

leosuncin:
Solves lorenwest/node-config#564

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1284,7 +1284,7 @@ util.isObject = function(obj) {
  * @returns {boolean}
  */
 util.isPromise = function(obj) {
-  return util.isObject(obj) && obj.toString() === '[object Promise]';
+  return util.isObject(obj) && Object.prototype.toString.call(obj) === '[object Promise]';
 };
 
 /**
--- a/lib/config.js
+++ b/lib/config.js
@@ -1284,7 +1284,7 @@ util.isObject = function(obj) {
  * @returns {boolean}
  */
 util.isPromise = function(obj) {
-  return util.isObject(obj) && Object.prototype.toString.call(obj) === '[object Promise]';
+  return Object.prototype.toString.call(obj) === '[object Promise]';
 };
 
 /**
--- a/test/util.js
+++ b/test/util.js
@@ -74,6 +74,8 @@ vows.describe('Tests for config util functions')
           new String(),
           [],
           {},
+          Object.create(null),
+          new Map(),
           null,
           undefined,
           NaN,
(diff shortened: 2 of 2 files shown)


leosuncin:
Call `toString` method from `Object.prototype` in case when the argument `obj`doesn't implement it

iMoses:
@leosuncin Can you provide a test with it to showcase the problem?

leosuncin:
I created a project with the error https://github.com/leosuncin/lorenwest-node-config-564

leosuncin:
I don't understand what do you ask @iMoses, can you clarify more or give me some examples?

iMoses:
@leosuncin If I understood #564 correctly then the problem looks something like that:
```js
const object = {a: 1};
object.prototype = null;
object.toString() // function doesn't exist
``` 
If our solution is supposed to solve this problem then I think we should have a test proving that it does.

I expected the test to look something like this:
```js
const object = {a: 1};
object.prototype = null;
assert(util.isPromise(object), false);
const promise = Promise.resolve({a: 1});
promise.prototype = null;
assert(util.isPromise(promise), true);
```

leosuncin:
Do you mean sometime like this?
> empties the object's prototype
```js
var obj = Object.create(null);
console.log(obj.__proto__, obj.constructor, obj.toString);
console.log(Object.prototype.toString.call(obj)); // Shows '[object Object]'
```
> empties the promise's prototype
```js
var promise = new Promise(() => {});
console.log(promise.__proto__, promise.constructor, promise.toString());
promise.__proto__ = null;
// from now any test will fail
console.log(promise.__proto__, promise.constructor, promise.toString);
// the previous test doesn't work
console.log(util.isObject(promise) && promise.toString() === '[object Promise]'); // <- Exception will throws
// neither my proposal
console.log(Object.prototype.toString.call(promise)); // Shows '[object Object]'
```
Because `promise.__proto__ = null` overrides the constructor name, this will never happens in normal code, so I think the last one is unnecessary test case

iMoses:
@leosuncin you may be right, but when it comes to tests (simple ones) I'd rather cover more cases.

If you asked me a week ago I wouldn't have thought empty prototypes is something I'd ever have to worry about..

leosuncin:
People will never override promise's `__proto__` because is going to break the promise:
```js
var promise = new Promise(function(resolve) {
  setTimeout(resolve('Hi from the future'), 2000);
});
console.log(promise.constructor.name, promise instanceof Promise, 'then' in promise, typeof promise.then);
promise.__proto__ = null;
console.log(promise.constructor, promise instanceof Promise, 'then' in promise, typeof promise.then);
promise.then(console.info); // Throws an exception
```

lorenwest:
Published today in config@3.2.1

Issue #521, opened 2018-12-15
Add esmodule support


asakurayoh1987:
When I write config file in .ts files it works well with ts-node in development environment, but after compiled it throws error with 'Configuration property "xxx" is not defined' , so I read the source and find the process with .js config files as bellow:

```javascript
if (extension === 'js') {
      // Use the built-in parser for .js files
      configObject = require(fullFilename);
}
````

but in the .ts file I use `export default xxx` and it compiled as bellow:
```javascript
Object.defineProperty(exports, "__esModule", { value: true });
...
exports.default = xxx
```

so, could you add support for es module just like bellow:  

```javascript
if (extension === 'js') {
    // Use the built-in parser for .js files
    configObject = require(fullFilename);
    // Process if the .js files is es module
    if(configObject.__esModule) {
        configObject = configObject.default;
    }
}
```

jasonwilliams:
I've hit the same issue, do you have a workaround @asakurayohj  ?

asakurayohj:
I just forked and  modified the code as above, then in the package.json use the git adress instead of the npm:
"config": "https://github.com/asakurayohj/node-config.git"

jasonwilliams:
@asakurayohj i managed to fix it by doing `import * as config from 'config';` in the .ts files

asakurayohj:
@jasonwilliams in defalut.ts (config file), I use `export default config`,when convert to js file, it changed into `exports.default = config;`, and in the source of lib config，just use `require(fullpath)`, it can not get the config object you want
could you paste the code

Kle0s:
I am still suffering from this issue. Has anyone managed to find a workaround?
I cannot use the library in an ES module I wrote in typescript

markstos:
@Kle0s I tested loading this module from a TypeScript file today and it worked fine:

```typescript
process.env.NODE_APP_INSTANCE='mytest';
process.env.NODE_ENV='laptop';

const config = require('config');

console.log('TEST', config.get('http.FQDN'));
```

We have supported config files in the .ts format format years.

This issue has wandered over the years. I'm closing it as stale. Open a new one with a clear title and description if there is more to discuss.

PR #566, opened 2019-07-16
Add ESModule support for compiled JS

ajcrowe:
This adds the same support for esmodule default exports to the `js` parsers that exists within the `ts` one already.

This resolves an isssue with compiled Typescript here: #521 
This was the original esmodules fix for the `ts` parser #545 

Another option for the condition would be checking `configObject.__esModule`

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -48,7 +48,16 @@ Parser.xmlParser = function(filename, content) {
 };
 
 Parser.jsParser = function(filename, content) {
-  return require(filename);
+  // Imports config if it is exported via module.exports = ...
+  // See https://github.com/lorenwest/node-config/issues/524
+  var configObject = require(filename);
+
+  // Because of ES6 modules usage, `default` is treated as named export (like any other)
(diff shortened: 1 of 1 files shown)


lorenwest:
I don't understand the problem that this is trying to address. It seems if someone has a `default` element in their config object, this will hide that key and expose only that element. 

Seems strange, and backwardly incompatible with existing configs if they happen to choose `default` as one of their config keys.

ajcrowe:
This was to solve issue where you use the  ES6 default export in Typescript such as

`export default config;`

Once compiled to javascript this then results in a need to reference the `default` export explicitly. Perhaps this isn't needed.

Perhaps checking `configObject.__esModule`  is a  better idea?

lorenwest:
It sounds safe to add if you check for `configObject.__esModule && configObject.default`. Make sure to include a test, maybe with a transpiled config file.

markstos:
Closing, as no test was ever provided. If someone wants this feature,  fork the work so far and add a test.

PR #568, opened 2019-07-20
Fix #567 - Missing path.delimiter breaks windows absolute paths

### LINKED ISSUE #567: Configuration files loading error on Windows
https://github.com/lorenwest/node-config/blob/ced8854a4bbf2ec09536d1db804d5cd5b60bb1ab/lib/config.js#L669
The hardcoded path delimiter (:) breaks handling of Windows file paths. OS dependent require('path').delimiter should be used.

iMoses:
I didn't add tests because we don't run tests on Windows servers as part of our CI.


lorenwest:
Published in `config@3.2.2`

PR #571, opened 2019-08-03
#569 Next-Gen: Imported new utilities and made raw obsolete

iMoses:
In an effort to reduce the amount of change with each push I started by importing only the new utilities method that should be internal anyhow. I replaced similar existing util methods with their new versions, which in turn made `raw()` obsolete as complex objects are protected by default.

```js
// next-gen utils
util.isObject = utils.isObject;
util.isPromise = utils.isPromise;
util.getCmdLineArg = utils.getArgv;
util.makeImmutable = utils.deepFreeze;
util.extendDeep = utils.extendDeep;
util.cloneDeep = utils.cloneDeep;
```

I moved `defer` and `async` inside `/lib` keeping a shim at the old locations, and replaced the old resolve logic with the new one (eventually completely similar). I replaced `raw()` with an identity method (`v => v`) and added a deprecation warning.

There are also two opinionated choices hiding in this commit.
- In case of duplicate cli-args we take the last instead of the first. This is common cli-args behavior
- When freezing the object we do not freeze dynamic getters values, but keep the getter functions

Not a lot of code has change in this commit it was mostly moved, though a bit more than is actually used was added (properties utilities are still unused). It will allow us to check at least some of the effects in advance of any major changes.

@lorenwest Do you support this commit?


lorenwest:
Yes I support this change - thank you for splitting it up. Give me a few days to download and review (i know - it doesn't take that long, but life...).

I'd like a separate thread on submodule support and if there's a possibility of introducing an alternate mechanism without breaking backward compatibility. First class submodule support allowed our company to take a modular approach to our large scale in-house applications, and is a game changer.

Other features can be discussed independently. For example I think it's cleaner to separate the config object from the config methods, but that breaks existing installs and when viewed independently, I'm not yet seeing the benefit outweighing the cost.

iMoses:
> For example I think it's cleaner to separate the config object from the config methods, but that breaks existing installs and when viewed independently, I'm not yet seeing the benefit outweighing the cost.

@lorenwest I think the benefits are huge...

The main change we'll have to make is to separate the actual configuration object from the Config instance which will enable us to lazy load files, making it possible to create multiple instance for package owners and define options (even parser options) from within NodeJS (instead of having to override env-vars as our users currently do).

Anything else can be postponed, although I believe that if we do add such a breaking change then we should go all the way to improve the API. Major versions are meant to be major and introduce breaking changes, and there's nothing wrong if some of our users will continue using version 3 if the changes are too much for them at the moment.

I've been using this library for a long while now and I do love it's simplicity and guidelines, but the API could really use a refactor. Too many patches have been added to a point that the code is unclear, there's lots of duplication of code, unused code, code that shouldn't be a part of the library at all (why do we have diff methods?).. And it's at a point that closing bugs which are open for several years now is nearly to impossible without making it even worst.

Let's work together to make sure that these changes are compatible with your company's work, or at least very easy to migrate, but if you look at the issues opened by the library's community then can you really claim that these changes are not worth it? Look at how many tickets there are about broken API methods. (setModuleDefault and immutability in general, using from packages, better support for parser and templating, etc)

lorenwest:
This is a good discussion - probably for another thread. I can see how multiple instances could be managed under the covers supporting submodules, lazy loading, and other code refactors while still maintaining a single config object outside the API. It may require the use of get() vs. direct access, but we've been recommending that path for a long time and it wouldn't be a bad compromise for users of new functionality. If the requirement is for new features only, it can't be considered backwardly incompatible.

Introducing breaking changes is easy - and sexy. Backward compatibility is hard work and comes with compromises, but the multiplier is huge vs. adding work for the users of the library.

iMoses:
@lorenwest I agree.. will talk about it in details on a dedicated thread.

BTW, we can keep most of the `config.util` methods, either as a indentical standalone or as an adapter for the new methods and add a deprecation warning to remove them somewhere in the future. The only actual breaking change that the entire next-gen idea is based on is the separation of the configuration object into a lazily loaded property. I'll give it some thought to see what more mid-stages I can poropose.

lorenwest:
I can imagine multiple objects, separated from code, lazily loaded all under the covers, all contributing to the single exposed object.

We may have to stop locking the top level in order to get lazily loaded submodule folding, but doing that wouldn't introduce work for the users. It may be considered incompatible because you can't rely on the top level being immutable, but compromises...

iMoses:
@lorenwest I see things differently, and I'd like the change to explain it in details over a call...

My proposal is to lock the config property on first access, therefore allowing any subModule the change to extend its values before this the first access. I'm in favor of keeping everything immutable and validating that it's stay this way, unless mutations are allowed, but it should work well with immutability too.

lorenwest:
The problem with submodule locking after the first get() is submodules are discovered well into the running of the application. You can force the main application to artificially load them all (if it knows them all - which it doesn't in a healthy modular system), or you can allow them to be discovered in due course and fold their configs into the main config after the first get().

I could imagine get() locking specific branches of the config object, allowing submodules to fold their namespace as long as nobody issued a get() for their config branch. This would force a require() of the submodule before using any of its configs, which shouldn't be a concern.

Let's take this to a new thread. There are a few major issues like this with submodules I'd like to address before proceeding.

markstos:
Good ideas and good work here thanks to @iMoses , but the branch is so stalled, I think any "next gen" discussions should be restarted in a new issue for those who wish to continue the discussion. One thing I see in the commits here is some changes to make sure old versions of Node.js are supported that we've since dropped support for some old Node versions. There were also be a number of conflicts to resolve between this branch and the main branch at this point.

This will serve as a valuable reference for anyone else considering a major update.

PR #576, opened 2019-10-01
FIX #460 - Strict mode warning refer to appropriate env variable

### LINKED ISSUE #460: Strict mode warning still refers to NODE_ENV variable even if overriding with NODE_CONFIG_ENV is used
When environment is overwritten via NODE_CONFIG_ENV, warning message should also consider this.

**Given:**
app is launched as 
`NODE_CONFIG_ENV=ut NODE_ENV=production node index.js`
no configuration file at path _config/ut.json_

**Actual message in console:**
`WARNING: NODE_ENV value of 'ut' did not match any deployment config file names.`

**Expected message:**
`WARNING: NODE_CONFIG_ENV value of 'ut' did not match any deployment config file names.`

iCodeOkay:
Fix for #460 .

Would love your feedback 🙂


markstos:
Looks OK to me. @lorenwest ?

PR #582, opened 2019-11-15
Allow all defined values in `substituteDeep`

fostyfost:


lorenwest:
Thank you for this contribution. It appears to be an improvement in that the current implementation doesn't allow all values (specifically, other falsey values like numeric 0 and the "" empty string).

I support this change, and since it will cause a slight behavior change I think it would be appropriate to add a test that outlines the problem. One that has a list of items to substitute including some non-undefined falsey values, and verifies that these values make their way into the substitution.

In the unlikely case that someone was relying on this side-effect, the test will show the problem clearly, and will be hard to argue that this shouldn't be considered correct behavior for substituteDeep.

markstos:
Looks good to me too.

fostyfost:
@lorenwest @markstos 
Hello, I added the tests you requested.

lorenwest:
Published today on `node-config@3.3.0`

wham:
This is definitely a change in the right direction. But it seems to be backwards incompatible and as such would deserve a major version update or at least a better description on the changelog. Given that environment variables are not strongly typed, I think it's a pretty common scenario to use `""` as _not set_.

PR #585, opened 2020-01-03
Fixed issue with getCustomEnvVars method and multiple config dirs

dekelev:
When working with multiple config dirs, custom-environment-variables files no longer works. 

The `getCustomEnvVars` method tries to create file paths by concatenating `CONFIG_DIR` with optional file names, but `CONFIG_DIR` can be a string with multiple dir paths, separated by `:`.


lorenwest:
This looks great - thanks for the addition. I'll get it merged and published shortly.

lorenwest:
Published today at config@3.2.5. Thanks for the contribution!

PR #586, opened 2020-01-07
Add support for experimental modules

Worthaboutapig:
ES Modules can be used in Node with the --experimental-modules flag.

However, when using the package.json "type": "module" option, which assumes all files in a package are ES modules, then using node-config fails, as the `require` in the parser is unexpected.

Renaming a non-module js file to .cjs allows the non-import usage. This PR adds support for a .cjs file extension for config files with format and behaviour identical to .js.

I appreciate that module support is experimental, however, it's fairly mature, it seems the API is unlikely to change and this change is backward-compatible.

Without this change node-config isn't really usable from a module-first nodejs approach, which is likely to be the default in future (unless I'm missing something).

(haven't added any tests, as not sure where to and you may not agree to the change anyway)

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -281,8 +281,9 @@ Parser.stripYamlComments = function(fileStr) {
   return fileStr.replace(/^\s*#.*/mg,'').replace(/^\s*[\n|\r]+/mg,'');
 };
 
-var order = ['js', 'ts', 'json', 'json5', 'hjson', 'toml', 'coffee', 'iced', 'yaml', 'yml', 'cson', 'properties', 'xml'];
+var order = ['js', 'cjs', 'ts', 'json', 'json5', 'hjson', 'toml', 'coffee', 'iced', 'yaml', 'yml', 'cson', 'properties', 'xml'];
 var definitions = {
+  cjs: Parser.jsParser,
   coffee: Parser.coffeeParser,
   cson: Parser.csonParser,
   hjson: Parser.hjsonParser,


lorenwest:
I'm not opposed to the change. Does someone who knows --experimental-modules better than myself chime in?

lorenwest:
I'd like to merge this. Would you be able to create a wiki page on this usage?

markstos:
Seems reasonable to me. @Worthaboutapig Are you willing to make a related wiki page update?

Worthaboutapig:
Hi, yeah, sure. Sorry, been a bit busy. Will get on to it.

Worthaboutapig:
> Seems reasonable to me. @Worthaboutapig Are you willing to make a related wiki page update?

Sure, should I do that before or after the merge?

Worthaboutapig:
Thanks. Am happy to update the docs. Should I do that directly, or is that a PR type process?

lorenwest:
Go ahead and add some documentation, stating "as of 3.4.0" and I'll merge and publish as 3.4.0. Let me know when you're ready and I'll publish.

Worthaboutapig:
> Go ahead and add some documentation, stating "as of 3.4.0" and I'll merge and publish as 3.4.0. Let me know when you're ready and I'll publish.

Sorry, could you clarify how to do this? I've cloned the wiki and updated, but I don't have permissions to push to the repo. Should I edit the pages directly though the website? I'm not familiar with wiki editing.

lorenwest:
Yes, the wiki is designed for community contribution and to be edited directly on the site. There should be an _Edit_ button on the page.

Worthaboutapig:
Sorry, I'll get around to this, been a bit busy.

JacobLey:
Ran into this issue as well, would love to see it unblocked.

Edited:
* .js section of https://github.com/lorenwest/node-config/wiki/Configuration-Files#commonjs-javascript-module---js-cjs
* intro of https://github.com/lorenwest/node-config/wiki/Special-features-for-JavaScript-configuration-files

markstos:
This seems ready to merge to me. There is a conflict, but it should be minor to resolve considering it's a 2 line change.

markstos:
Resolved by #622.

PR #587, opened 2020-01-14
Run ts-node in mode transpile-only

bvallee-thefork:
By default, ts-node is type checking the imported files, doing so on TypeScript config files creates a requirement to have `@types/config` in `dependencies` (as opposed to `devDependencies`) when `config/defer` (or similar) is imported from one of the configuration files.

To read the configuration values, only `transpile-only` is needed (which is also faster to run).

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -55,6 +55,7 @@ Parser.tsParser = function(filename, content) {
   if (!require.extensions['.ts']) {
     require(TS_DEP).register({
       lazy: true,
+      transpileOnly: true,
       compilerOptions: {
         allowJs: true,
       }


lorenwest:
Could someone using ts-node chime in here? I don't know it well enough to know if this has backward compatibility issues.

bvallee-thefork:
> Could someone using ts-node chime in here? I don't know it well enough to know if this has backward compatibility issues.

It shouldn't have any backward compatibility issues, basically it does the same thing as before but without running all the types checking from from `tsc`/`ts-node` (no difference at runtime).

I let someone else jump in for the purpose of peer-review :)

lorenwest:
Ping me after a while, and if nobody jumps in I'll go ahead and merge.

bvallee-thefork:
Ping @lorenwest ? :)

lorenwest:
Does this turn off type checking? If so, why use TypeScript?

bvallee-thefork:
> Does this turn off type checking? If so, why use TypeScript?

It turns off type checking at runtime, the type checking should be done either at build time or at lint stage (which also allows `@types/` packages to be in `devDependencies`, but is not possible when `ts-node` is run with `transpileOnly: false`).

lorenwest:
@benedyktdryl added TS functionality. Would you mind chiming into this thread? Is this change the way it was originally intended, or is this a fundamental shift from the original intent?

benedyktdryl:
Shouldn't be a problem but what is point of using TS compilator on library side if you are not using type checking?

bvallee-thefork:
> Shouldn't be a problem but what is point of using TS compilator on library side if you are not using type checking?

@benedyktdryl When running code with `ts-node`, it's a good practice to have type checking done at lint stage, so it doesn't run extra type checking at runtime, and because ts-node doesn't support all types definition features from `tsconfig.json`.

PR #590, opened 2020-01-30
Add Ability to use Hostnames with Local Configuration Files

seesemichaelj:
Hello!

My company found the temporary need to have `local.EXT` configuration files that were augmented by the hostname. I created the work to support this and used it within development, but now no longer need it. However, I figured that the work shouldn't go to waste if it's something you @lorenwest would like to incorporate into the upstream fork.

I could try to squeeze some time to add tests for this, but not sure when I'd be able to get around to that.

We love your library; thank you for saving me and others so much time with this awesome library (we had a haphazard `.env` setup for our large multi-environment monorepo)

Let me know what your thoughts are! Happy to answer any questions. Basically, `local-(hostname).EXT` and `local-(hostname)-(deployment).EXT` are added to the end of the order of superseding config.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -583,20 +583,23 @@ util.loadFileConfigs = function(configDir) {
     });
   }
 
+  baseNames.push('local');
   NODE_ENV.forEach(function(env) {
-    baseNames.push('local', 'local-' + env);
+    baseNames.push('local-' + env);
   });
 
 
   if (firstDomain) {
+    baseNames.push('local-' + firstDomain);
     NODE_ENV.forEach(function(env) {
-      baseNames.push('local-' + firstDomain, 'local-' + firstDomain + '-' + env);
+      baseNames.push('local-' + firstDomain + '-' + env);
     });
   }
 
   if (hostName && hostName !== firstDomain) {
+    baseNames.push('local-' + hostName);
     NODE_ENV.forEach(function(env) {
-      baseNames.push('local-' + hostName, 'local-' + hostName + '-' + env);
+      baseNames.push('local-' + hostName + '-' + env);
     });
   }
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -459,6 +459,8 @@ util.getConfigSources = function() {
  *   (hostname)-(deployment).EXT
  *   local.EXT
  *   local-(deployment).EXT
+ *   local-(hostname).EXT
+ *   local-(hostname)-(deployment).EXT
  *   runtime.json
  * </pre>
  *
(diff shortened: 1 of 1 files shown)


lorenwest:
I'd like to understand your use case before considering this for merge. 

local.EXT was designed to be created at deploy time, and checked in to the VCS. In this context, the local.EXT file can't clash because by definition it's a file created for this hostname/deployment.

seesemichaelj:
First, we don't check in `local*.EXT` to VCS; we use it as a development override. One member of the team uses different GitHub api keys than another to prevent webhooks from being received to both developers, and we use `local*.EXT` for this. Perhaps we're misusing it than intended, but it works really well for us.

The use case of adding hostname into the `local*.EXT` chain was that I have a monorepo of multiple packages/microservices which shares the same `config` directory as we use some configuration across the entire application (i.e. where the database is hosted, and API key that multiple services use, etc). In development, I required to override some pieces of the configuration for different microservices.

The specific use case I ran into: I have a frontend client-side-only application that connects to some server. The backend services also connect to the same server. In development, I wanted to selfhost that server rather than pointing to a public API of another service that provides that server. The backend services needed to point to another docker microservice hostname `my-server:port`, but my frontend application runs code in the browser locally, so it doesn't have access to the docker container `my-server`. To resolve this, `my-server` was configured to expose the port to `localhost` and I could configure the front end application to point to `localhost:port` instead of `my-server:port`

Now typing it out, it does sound like an edge case haha. I'm totally fine if it doesn't make sense to be merged upstream, but wanted to provide it as an option to you. I can keep it in my personal fork if I ever desire the need to reuse it.

lorenwest:
The local.EXT is designed to be unique to a host, so solving for multiple hostnames is counter-intuitive to the purpose of local.EXT.  Thank you for offering this PR, but I think it's best to pass on merging to prevent confusion.

seesemichaelj:
Sounds good!

markstos:
In this case, if there was a "local.EXT" file inside the docker container then it would be isolated from from the backend service running outside the container, so two different local.EXT files could be used on the same host?

Ah, but I think you are saying they are both in the monorepo and would normally share the same config files... the difference is that the Docker container has a different host name than the service running outside of it. 

If you are using "mount" inside the docker container now, you could use ADD or COPY directories to copy a file into the docker container. Another line of the docker configuration could modify the local.EXT file as needed. 

There are variations on that theme, but I think there's a solution here that involves updating how the Docker image is built so that it gets the unique configuration that it needs.

seesemichaelj:
>There are variations on that theme, but I think there's a solution here that involves updating how the Docker image is built so that it gets the unique configuration that it needs.
Totally! I mean, I have the solution I made, and I can always use my fork whenever I need it (which is not often), so I really don't need a solution anymore. Thanks though!

PR #594, opened 2020-03-02
Ability to parse environment variables as number or boolean

cur3n4:
Related to #589, this pull requests allow to specify a custom variable to be formatted via boolean or number.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -199,6 +199,14 @@ Parser.propertiesParser = function(filename, content) {
   return PPARSER.parse(content, { namespaces: true, variables: true, sections: true });
 };
 
+Parser.numberParser = function(filename, content) {
+    return Number(content);
+}
+
+Parser.booleanParser = function(filename, content) {
+    return ('' + content === 'true');
+}
+
 /**
  * Strip all Javascript type comments from the string.
  *
@@ -281,7 +289,7 @@ Parser.stripYamlComments = function(fileStr) {
   return fileStr.replace(/^\s*#.*/mg,'').replace(/^\s*[\n|\r]+/mg,'');
 };
 
-var order = ['js', 'ts', 'json', 'json5', 'hjson', 'toml', 'coffee', 'iced', 'yaml', 'yml', 'cson', 'properties', 'xml'];
+var order = ['js', 'ts', 'json', 'json5', 'hjson', 'toml', 'coffee', 'iced', 'yaml', 'yml', 'cson', 'properties', 'xml', 'number', 'boolean'];
 var definitions = {
   coffee: Parser.coffeeParser,
   cson: Parser.csonParser,
@@ -296,6 +304,8 @@ var definitions = {
   xml: Parser.xmlParser,
   yaml: Parser.yamlParser,
   yml: Parser.yamlParser,
+  number: Parser.numberParser,
+  boolean: Parser.booleanParser
 };
 
 Parser.getParser = function(name) {
--- a/test/2-config-test.js
+++ b/test/2-config-test.js
@@ -14,7 +14,7 @@ var vows = require('vows'),
  * @class ConfigTest
  */
 
-var CONFIG, MODULE_CONFIG, override;
+var CONFIG, MODULE_CONFIG, override, overrideBoolean, overrideNumber;
 vows.describe('Test suite for node-config')
 .addBatch({
   'Library initialization': {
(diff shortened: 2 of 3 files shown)


lorenwest:
This looks good to me, but I'd like someone that uses custom environment variables to chime in

markstos:
I use custom environment variables.

I feel mixed about this. 

On one hand, it is nice to convert strings to booleans within one place in the config system. The user of the `config` library shouldn't necessarily be concerned if the config value was set through the environment (which only supports strings) or in a config file where booleans and numbers are natively supported.

On the other hand, it's easy to use a module like `to-boolean`, which handles exactly this task. It supports `y`, `1`, `yes`, and `true` for truthy values. It's not clear that this patch supports all these cases.  https://github.com/rafaelrinaldi/to-boolean#readme

In the end I think I learn towards accepting this, but have it support all the common shorthands for booleans as noted above.

parser.js
```diff
@@ -199,6 +199,14 @@ Parser.propertiesParser = function(filename, content) {
   return PPARSER.parse(content, { namespaces: true, variables: true, sections: true });
 };
 
+Parser.numberParser = function(filename, content) {
+    return Number(content);
+}
+
+Parser.booleanParser = function(filename, content) {
+    return ('' + content === 'true');
+}
```
marlomajor, line 208: Change both of these to:
```
Parser.booleanParser = (filename, content) => {
    return ('' + content === 'true');
}
```

nflaig:
I agree with @markstos, it would be really good if all common boolean shorthands could be used. I would recommend using [yn](https://github.com/sindresorhus/yn) for this purpose.

Also one more really handy thing would be to have the option to set format to `time` which would allow values like "2 days" or "5h". This could be also easily achieved by using a library such as [zeit/ms](https://github.com/zeit/ms).

I know this comes with the expense of adding additional dependencies but in my opinion would be worth it.

lorenwest:
I could be convinced to support type conversion to match the functionality provided by JSON, but I'd draw the line at more than type conversion. `y` means `"y"`, not boolean `true`.  If you were expecting a boolean and someone put a `"y"` in their default.json file, you'd have a problem.

Is there a reason that this [existing solution](https://github.com/lorenwest/node-config/issues/589#issuecomment-596393370) isn't appropriate?

nflaig:
I would also say that it is sufficient. 

The issues mentioned in [#589 ](https://github.com/lorenwest/node-config/issues/589) are solved by this PR.

parser.js
```diff
@@ -199,6 +199,14 @@ Parser.propertiesParser = function(filename, content) {
   return PPARSER.parse(content, { namespaces: true, variables: true, sections: true });
 };
 
+Parser.numberParser = function(filename, content) {
+    return Number(content);
```
SkeLLLa, line 203: What value will be in result config in case if we have empty env variable?
```
MY_VAR=
```

```
"number": {
            "__name": "MY_VAR",
            "__format": "number"
        }
```
And in default.js
```
module.exports = {
  number: 1
}
```

fiznool: Quirky JavaScript:

```
> Number(undefined)
NaN
> Number("")
0
> Number(null)
0
> Number("1")
1
> Number("0.5")
0.5
```

I'd suggest an extra condition here to account for the `undefined` case, since this is what will happen if the env var is not set:

``` js
Parser.numberParser = function(filename, content) {
  const val = Number(content);
  return Number.isNaN(val) ? undefined : val;
};
```

parser.js
fiznool: Given that the rest of this file uses `function()` instead of `() =>` I'd suggest leaving it as it is.

fiznool:
This looks good to me. One big advantage over the `__format: 'json'` solution is that it handles the absence of an environment variable gracefully, whereas using the JSON formatter, you end up with `JSON.parse(undefined)` which throws an exception.

What can we do to move this forward? I notice an absence of tests, which would be critical IMHO to verify the correct behaviour. I'd love to see this feature in place!

marcoreni:
Any updates on this?

markstos:
Please rebase to resolve conflicts and a comment here that is a preview of docs that could be added to the wiki to document this feature.

nflaig:
@markstos the same changes are already merged from a different PR, see #616 which resolved #589

PR #606, opened 2020-05-26
Support .cjs files as alias for .js

JacobLey:
Hello, I normally use `.js` files for writing my config files. I have been trying to migrate running native ESM, and noticed that breaks the current parser logic.

_Ideally_ I would love to write ESM syntax for configs as well, but considering these files are loaded dynamically __and__ synchronously, that would require some new parser. I'd rather not deal with that for now.

Instead, I'd like to just migrate my config files to `.cjs` and maintain the existing loading pattern.

I could open a bug/feature issue to further discuss this if necessary, but seems pretty straightforward to me.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -281,8 +281,9 @@ Parser.stripYamlComments = function(fileStr) {
   return fileStr.replace(/^\s*#.*/mg,'').replace(/^\s*[\n|\r]+/mg,'');
 };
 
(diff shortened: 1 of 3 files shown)


JacobLey:
🤦 I see https://github.com/lorenwest/node-config/pull/586 as well.

Seems like the blocker is the wiki? I can take a swing at that

JacobLey:
Edited:
* .js section of https://github.com/lorenwest/node-config/wiki/Configuration-Files#commonjs-javascript-module---js-cjs
* intro of https://github.com/lorenwest/node-config/wiki/Special-features-for-JavaScript-configuration-files

markstos:
Closing as dupe of #586.

PR #608, opened 2020-05-29
Fixed issue with Buffers in config throwing error in util.makeImmutable

Ginden:
Simple code replicating issue:

`config/default.js`
```javascript
module.exports = {foo: Buffer.from('bar', 'ascii'), bar: 5};
```
`index.js`
```js
require('config').get('bar')
```

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -369,6 +369,9 @@ util.makeHidden = function(object, property, value) {
  * @return object {object} - The original object is returned - for chaining.
  */
 util.makeImmutable = function(object, property, value) {
+  if (Buffer.isBuffer(object)) {
+    return object;
+  }
   var properties = null;
 
   // Backwards compatibility mode where property/value can be specified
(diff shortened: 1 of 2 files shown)


markstos:
This looks good, but since Buffer values are uncommon, let's put the check towards the end the chain of checks in makeImmutable, so we aren't checking every single value to see if it's a Buffer when most of them won't be.

Ginden:
I don't think [single `instanceof` check](https://github.com/nodejs/node/blob/9cd83c761f688079efeae931cb3462b9ba6d18e3/lib/buffer.js#L520) matters for performance of cold function. 

On my PC checking **one million elements** take around 100ms.

markstos:
Ha, good enough. Merging.

ekoeryanto:
need it soon

lorenwest:
Can't believe this hasn't been published - apologies for it taking so long!  Published in 3.3.2

PR #611, opened 2020-06-18
Add NODE_CONFIG_RUNTIME process.env variable to completely overwrite all config data

suil:
process.env.NODE_CONFIG_RUNTIME runtime variable can only overwrite on top of config data after being loaded from files. 

We need a process.env variable that can completely bypass config files loading process. So that we could create json file in the build time and only upload compiled config data to servers.

Also with new process.env.NODE_CONFIG_RUNTIME variable, we could use `webpack.DefinePlugin` to ship bundled artifact without copy all files under config folder.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -16,7 +16,7 @@ var deferConfig = require('../defer').deferConfig,
 var DEFAULT_CLONE_DEPTH = 20,
     NODE_CONFIG, CONFIG_DIR, RUNTIME_JSON_FILENAME, NODE_ENV, APP_INSTANCE,
     HOST, HOSTNAME, ALLOW_CONFIG_MUTATIONS, CONFIG_SKIP_GITCRYPT,
-    NODE_CONFIG_PARSER,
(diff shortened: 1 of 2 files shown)


suil:
I found some tests are failing. After you agree with my approach, I will add tests and modify readme.

markstos:
I'm not sure I agree this is needed. If your build system is already modifying things, why not just delete the config files that you don't want to ship to production?

 I do that for a project and it's easy enough. For example, for a server in Canada, I don't ship the configs that only apply to customers in the United States.

markstos:
> process.env.NODE_CONFIG_RUNTIME runtime variable can only overwrite on top of config data after being loaded from files.

Do you actually intend to refer to `RUNTIME_JSON_FILENAME` here?

This is a challenge with this feature proposal-- there's already an environment variables with a similar name that does a slightly different thing.

markstos:
Closing because this sounds like what `NODE_CONFIG` already does.  

https://github.com/node-config/node-config/wiki/Environment-Variables#node_config

PR #640, opened 2021-02-23
introduce an optional options object to the parseFile and loadFileCon…

### LINKED ISSUE #517: loadFileConfigs incorrectly adds to getConfigSources
### I'm submitting a ... 
  - [x] bug report

The documentation for `getConfigSources` says:

> The `config.util.getConfigSources();` method can be used to see all sources contributing to the ultimate configuration, and the order in which they were are applied.

However, this isn't accurate if you call `loadFileConfigs`:

```js
const otherConfig = config.util.loadFileConfigs('./other-config-dir');

config.util.getConfigSources().forEach(source => console.log(` - ${source.name}`));
```

Prints:

```
 - [...]/demo-app/config/default.js
 - [...]/demo-app/config/development.js
 - [...]/demo-app/other-config-dir/default.js
```

This is inaccurate, as `otherConfig` hasn't been merged at all with `config` yet, so it is not a source for what is contained in `config`. It hasn't affected `config` in any way, only `otherConfig`.

If you then use this value to set defaults via `setModuleDefaults`, it's still inaccurate, printing:

```
 - Module Defaults
 - [...]/demo-app/config/default.js
 - [...]/demo-app/config/development.js
 - [...]/demo-app/other-config-dir/default.js
```

Now that file is technically a source, but it's in the wrong spot – its contents are actually part of Module Defaults (coming first), the opposite of what is implied by the docs (they are not the final override in the cascade).

Configs loaded with `loadFileConfigs` should not appear in `getConfigSources`.

Potentially, the methods that one can use to merge config data (e.g. `extendDeep`, `setModuleDefaults`) could take a special argument or field (using Symbol?) for the developer to specify the source the config data came from, in which case they could safely appear there in the correct order no matter how the developer merges them in.

 -  node-config version: 3.0.0
 -  node-version: 10.13.0
 
### Other information

(e.g. are you using an environment besides Node.js?)

NguyenMatthieu:
…figs utility function, supporting a skipConfigSources flag, to fix #517

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -448,6 +448,29 @@ util.getConfigSources = function() {
 };
 
 /**
+ * Looks into an options object for a specific attribute
+ *
+ * <p>
+ * This method looks into the options object, and if an attribute is defined, returns it,
+ * and if not, returns the default value
+ * </p>
+ *
+ * @method getOption
+ * @param options {Object | undefined} the options object
+ * @param optionName {string} the attribute name to look for
+ * @param defaultValue { any } the default in case the options object is empty, or the attribute does not exist.
+ * @return options[optionName] if defined, defaultValue if not.
+ */
+util.getOption = function(options, optionName, defaultValue) {
+  if (options !== undefined && options[optionName] !== undefined){
+    return options[optionName];
+  } else {
+    return defaultValue;
+  }
+};
+
+
+/**
  * Load the individual file configurations.
  *
  * <p>
@@ -635,7 +658,7 @@ util.loadFileConfigs = function(configDir, options) {
         console.error('The $NODE_CONFIG environment variable is malformed JSON');
       }
       util.extendDeep(config, envConfig);
-      var skipConfigSources = options !== undefined && options.skipConfigSources === true;
+      var skipConfigSources = util.getOption(options,'skipConfigSources', false);
       if (!skipConfigSources){
         configSources.push({
           name: "$NODE_CONFIG",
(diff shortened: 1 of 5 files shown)


lib/config.js
```diff
@@ -812,8 +821,9 @@ util.parseFile = function(fullFilename) {
     throw new Error("Cannot parse config file: '" + fullFilename + "': " + e3);
   }
 
-  // Keep track of this configuration sources, including empty ones
-  if (typeof configObject === 'object') {
+  // Keep track of this configuration sources, including empty ones, unless the skipConfigSources flag is set to true in the options
+  var skipConfigSources = options !== undefined && options.skipConfigSources === true;
```
lorenwest: This pattern is repeated several times, and given the nature of the optional options object it's likely it will be repeated in the future.  Do you think this could be condensed into a method that handles these edge cases and can be maintained in one place?

NguyenMatthieu:
Sure. What kind of signature do you have in mind?

lib/config.js
NguyenMatthieu: @lorenwest Sure. What kind of signature do you have in mind? :)

lorenwest: Sure, something like this:
```
const skipConfigSources = getOption('skipConfigSources', true);
```
The method could be defined within `loadFileConfigs()` for now

NguyenMatthieu: All right, I'll get started on it, shouldn't take long. Since it's used in loadFileConfigs *and* in parseFile, I'll put it a bit higher, though.
getOption(optionName: string, defaultValue) ?

lorenwest: Sounds good, and since it's outside of the function you'll have to pass the options object as arg1.
```
getOption(optionsObj: object, optionName: string, defaultValue: any);
```
make sense?

NguyenMatthieu: obviously :D.

NguyenMatthieu: @lorenwest done :)

lorenwest:
Looks great - thank you for the contribution!

PR #645, opened 2021-03-31
fix(parser): VisionmediaYaml.eval is not a func

SyMind:


SyMind:
@lorenwest

lorenwest:
Hi @SyMind,

This code appears to be very old.  Do you know if there was a version of the yaml parser that changed to remove the eval function?  If someone has the older version, will this fail?  

To keep this library as lightweight as possible we don't include dependencies, and the unfortunate side effect is it relies on the person using the library to install compatible versions.

If it is what I suspect (eval used to be there, but is no longer there), then the proper fix would be to test for the existence of the eval function, and use it if present, otherwise use the parse function.

Make sense?

SyMind:
@lorenwest You are right. `yaml` used `eval` method in 0.x.x version, used `parse` method in newer version. I changed the code to test the eval is a function.

lorenwest (approved):

markstos:
Before we release this change, I suggest adjusting the logic. As it stands, the `eval` method was replaced with the `parse` method in the YAML package over 3 years ago and there's no indication that it's coming back.

But in this new code, a check for the `eval` method happens every single time, even though the number of people using that code path is near-zero now and will be quickly trending towards zero.

I recommend instead directly calling `parse()`, but in a `try` block. If that fails because the `parse()` method doesn't exist, re-try the `eval()` method. That design flips the default, causing the extra work to happen for the people who have an ancient YAML library, not the ones who have a current one.

markstos:
I'm also applying the 4.0 label to this. In the next breaking change release, we should consider dropping support for the ancient YAML library completely. I'm OK with doing that breaking change sooner rather than later. It's been 3 years since the API was changed.

PR #649, opened 2021-04-22
Fix #593 (Don't treat empty strings as valid values)

### LINKED ISSUE #593: empty environment variables don't read default
### I'm submitting a ... 
  - [x] bug report
  - [ ] feature request
  - [ ] support request or question => Please do not submit support request or questions here, see note at the top of this template.

In node-config before 3.3.0 if environment variables is empty read from default but in 3.3.0 if environment variables is empty (like: `MY_ENV= node index.js`) don't read default

In version 3.2.6:

```bash
SERVER_HOST= HTTP_PORT= node index.js

### Output
### { host: '0.0.0.0', http: { port: 3000 }, https: { port: 3443 } }
```

In version 3.3.0:

```bash
SERVER_HOST= HTTP_PORT= node index.js

### Output
### { host: '', http: { port: '' }, https: { port: 3443 } }
```

My config dir info:

- config/default.json

```json
{
  "server": {
    "host": "0.0.0.0",
    "http": {
      "port": 3000
    },
    "https": {
      "port": 3443
    }
  }
}
```

- config/custom-environment-variables.json

```json
{
  "server": {
    "host": "SERVER_HOST",
    "http": {
      "port": "HTTP_PORT"
    },
    "https": {
      "port": "HTTPS_PORT"
    }
  }
}
```

 -  node-config version: 3.3.0
 -  node-version: 12.13.1

airdrummingfool:
This is a first-attempt PR to add tests to demonstrate the problem described in #593, and to fix that issue. It keeps the behavior that allows other falsey values, such as numeric `0`, which was a main benefit of the PR (#582) that stopped ignoring empty environment variables.

Note that I had to adjust a couple existing tests that broke due to my fix; I believe the tests were describing behavior that is incompatible with ignoring empty strings, and both tests were added at the same time or after the old behavior was broken in 3.3.0. Please let me know if this is incorrect or undesired.

Also note that I am unfamiliar with the test format that `node-config` uses, so I'd be happy to reformat/reorganize my added tests if there's a better place to put them - just let me know.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1084,12 +1084,12 @@ util.substituteDeep = function (substitutionMap, variables) {
     for (var prop in map) {
       var value = map[prop];
       if (typeof(value) === 'string') { // We found a leaf variable name
-        if (vars[value] !== undefined) { // if the vars provide a value set the value in the result map
+        if (vars[value] !== undefined && vars[value] !== '') { // if the vars provide a value set the value in the result map
           util.setPath(result, pathTo.concat(prop), vars[value]);
         }
       }
       else if (util.isObject(value)) { // work on the subtree, giving it a clone of the pathTo
-        if ('__name' in value && '__format' in value && vars[value.__name] !== undefined) {
+        if ('__name' in value && '__format' in value && vars[value.__name] !== undefined && vars[value.__name] !== '') {
           try {
             var parsedValue = util.parseString(vars[value.__name], value.__format);
           } catch(err) {
--- a/test/1-protected-test.js
+++ b/test/1-protected-test.js
@@ -347,7 +347,6 @@ vows.describe('Protected (hackable) utilities test')
       assert.deepEqual(substituted, {
         TopLevel: 0,
         Customers: {
-          dbName: '',
           oauth: {
             key: 'null',
             secret: 'false'
--- a/test/19-config/custom-environment-variables.json
+++ b/test/19-config/custom-environment-variables.json
@@ -1,0 +1,7 @@
+{
+  "testValue": "TEST_VALUE",
+  "testJSONValue": {
+    "__name": "TEST_JSON_VALUE",
+    "__format": "json"
+  }
+}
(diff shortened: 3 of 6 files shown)


fostyfost:
Why do you think empty var is invalid value?

airdrummingfool:
They're invalid in environment variable overrides, because of the behavior described in #593. Specifically, empty environment variables (which are read as empty strings) should not override config values set elsewhere, per this snippet of the [wiki](https://github.com/lorenwest/node-config/wiki/Environment-Variables#custom-environment-variables):

> Empty environment variables are ignored, and their mappings have no effect on your config.

They were previously (before 3.3.0) treated as invalid everywhere, so this PR just reverts that part of the behavior. Perhaps a future enhancement could be to allow them in places other than environment variables, but for now this is a partial reversion of the behavior change in #582.

markstos:
@lorenwest I'm in favor of merging this PR.

I believe it's correct to ignore empty-string environment variables. We can't tell if their intent is to set no value or intentionally set an empty value, so we have choose one expectation and document that, which has already been done. 

While some developers are unaware, environment variables are always converted to strings by Node.js. While within Node.js you can set `process.env.FOO = null` or `process.env.FOO = undefined`, the only thing that's possible in the environment that's passed to Node.js is `FOO=` or `FOO=''`, and those will either come through as empty string or not at all.

lorenwest:
I'm ok with it, as it appears to be a compatibility fix.  3.3.0 was not intended to break compatibility, and this resolves that problem.

airdrummingfool:
@lorenwest @markstos is there anything I can do to improve this PR or help it move forward?

markstos:
I think it's just waiting on @lorenwest to merge and release.

markstos:
Merged. Usually @lorenwest does the releases.

snewcomer:
@lorenwest Would it be possible to publish a `3.4.0` (or `3.3.7` release) that would include this fix when you have a chance?

Issue #660, opened 2021-09-28
NODE_CONFIG_DIR multiple relative directory paths separated by path.delimiter throws an error


inside:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `config@3.3.6` for the project I'm working on.

### I'm submitting a ... 
  - bug report

### Reproduction repository

https://github.com/inside/config-multiple-directories

### Related issues

https://github.com/lorenwest/node-config/issues/632

### What is the current behavior?

When assigning multiple relative directory paths to NODE_CONFIG_DIR, the following error is thrown:
```
/home/inside/src/test-config2/node_modules/config/lib/config.js:844
    throw new Error("Cannot parse config file: '" + fullFilename + "': " + e3);
    ^

Error: Cannot parse config file: 'configB/default.js': Error: Cannot find module 'configB/default.js'
Require stack:
- /home/inside/src/test-config2/node_modules/config/parser.js
- /home/inside/src/test-config2/node_modules/config/lib/config.js
- /home/inside/src/test-config2/index.js
    at Config.util.parseFile (/home/inside/src/test-config2/node_modules/config/lib/config.js:844:11)
    at /home/inside/src/test-config2/node_modules/config/lib/config.js:644:26
    at Array.forEach (<anonymous>)
    at Config.util.loadFileConfigs (/home/inside/src/test-config2/node_modules/config/lib/config.js:643:16)
    at new Config (/home/inside/src/test-config2/node_modules/config/lib/config.js:116:27)
    at Object.<anonymous> (/home/inside/src/test-config2/node_modules/config/lib/config.js:1492:31)
    at Module._compile (internal/modules/cjs/loader.js:1072:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1101:10)
    at Module.load (internal/modules/cjs/loader.js:937:32)
    at Function.Module._load (internal/modules/cjs/loader.js:778:12)
```

### What is the expected behavior?

The config should get multiple directories regarldess of their relative or absolute path

### Please tell us about your environment:

 -  node-config version: 3.3.6
 -  node-version: 14.17.5

### Here is the diff that solved my problem:

```diff
diff --git a/node_modules/config/lib/config.js b/node_modules/config/lib/config.js
index 0640cf5..d353248 100644
--- a/node_modules/config/lib/config.js
+++ b/node_modules/config/lib/config.js
[...]

PR #661, opened 2021-10-08
Allow multiple relative directory paths separated by path.delimiter to work

inside:
- Add tests for the multiple configurations feature
- transform each relative config dir to an absolute path

fix: https://github.com/lorenwest/node-config/issues/660

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -563,9 +563,7 @@ util.loadFileConfigs = function(configDir, options) {
   NODE_ENV = NODE_ENV.split(',');
 
   CONFIG_DIR = configDir || util.initParam('NODE_CONFIG_DIR', Path.join( process.cwd(), 'config') );
-  if (CONFIG_DIR.indexOf('.') === 0) {
-    CONFIG_DIR = Path.join(process.cwd() , CONFIG_DIR);
-  }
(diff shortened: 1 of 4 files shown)


lib/config.js
```diff
@@ -1488,6 +1487,15 @@ util.runStrictnessChecks = function (config) {
...
 
+// Helper functions shared accross object members
+function _toAbsolutePath (configDir) {
+  if (configDir.indexOf('.') === 0) {
+    return Path.join(process.cwd(), configDir);
+  }
+
+  return configDir;
+}
+
```
markstos, line 1498: Is there a reason not to use the built-in `path.resolve()` here?

https://nodejs.org/api/path.html#path_path_resolve_paths

inside: Using the original implementation (== using `process.cwd()`) is to not break the code.
`path.resolve()` seems appropriate for this case. If it's worth it should probably be done in another PR. What do you think?

markstos: I see now how we we using `process.cwd()` instead of `path.resolve()`, and agree that change is best made in another PR, so I'm merging this.

inside: Thank you!

PR #664, opened 2021-11-09
fix: Parser.jsParser fix

stopsopa:
fix for case when NODE_CONFIG_DIR="environment" - relative path without dots.

It's still valid relative path and it should be supported. 

It was working for a while in our project until we have introduced environment/custom-environment-variables.js then Parser.jsParser was confused calling internally require('environment/custom-environment-variables.js')

if NODE_CONFIG_DIR will be specified as relative path (without dots) then Parser.jsParser might encounter paths like 'environment/custom-environment-variables.js' and loading such path directly to require('environment/custom-environment-variables.js') will be interpreted as loading file 'custom-environment-variables.js' from 'environment' library.. which in most case might not be what you want.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -562,10 +562,7 @@ util.loadFileConfigs = function(configDir, options) {
   // Split files name, for loading multiple files.
   NODE_ENV = NODE_ENV.split(',');
 
-  CONFIG_DIR = configDir || util.initParam('NODE_CONFIG_DIR', Path.join( process.cwd(), 'config') );
-  if (CONFIG_DIR.indexOf('.') === 0) {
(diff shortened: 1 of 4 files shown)


markstos:
This is not a fix, the software is working as documented, which is that a leading dot is required for relative paths when using this variable:

https://github.com/lorenwest/node-config/wiki/Configuration-Files#config-directory

stopsopa:
So let me recap:
If developer will follow his life experiance thinking that "environment" and "./environment" are equally valid relative paths and choose form "environment" the library will work 95% of time for all combination. 

But if at some point he realise that you need environment/custom-environment-variables.js file then library will stop working...

...and that's ok because that's how it is documented?

Well that's your library...

PR #667, opened 2022-01-15
Cleanup of file scoped environment variables

jdmarshall:
This change removes all of the unused variables and fixes one case of variable shadowing.

I am slowly working toward separating config.util into its own file, but there are a number of couplings between it and the parent scope that will need to be sorted out.


markstos:
These all look good. Thanks!

Issue #572, opened 2019-08-04
[feature] Improved submodule support


lorenwest:
Let's use this thread to host issues related to upgrading submodule support. We should start by listing the functionality we'd like to see.  I'll start:

* Dual purpose - Submodules run both as submodules and as first class applications. Mostly they run as apps for testing, but they must be able to run in either mode. When running in application mode, all configurations at the top level of their config object are used. When running in submodule mode, only the top level config branch matching the module name is used.

* Submodule config as default - When loaded as a submodule, a full config object should be loaded from files in the /config directory of the submodule. Once loaded, the top level branch of that config object named by the module name is folded as defaults into the same top branch of the main app config.

* Late loading - A healthy modular system doesn't require the main app to have direct knowledge of all modules in the dependency graph. The main app knows only of dependencies listed in package.json. It should not be forced to change as the dependency graph changes over time. In addition to this lack of knowledge, the main app doesn't control the timing of submodule loading. Submodule configs must be able to be loaded as the runtime discovers the need to load the module.

* Backward compatibility - I'm listing this as a highly desired but optional feature. This feature may force us to hide the config object, requiring all access through `get()` in order to guarantee immutability.

lorenwest:
Ideally we can establish a pattern that doesn't classify a module as *submodule* - where all modules using node-config work in both modes.

lorenwest:
We must address the reality that some submodules will be loaded multiple times with different versions. We can either:

1) Try to address the complexity of multiple config values for different submodule versions, or
2) Allow only one version - first loaded wins, then gets merged into the top level app config

I fear that if we support #1, this will make app level overrides impossible (which version is the app overriding?), leaving #2 the only viable option - unless there's a #3 I'm not seeing.

iMoses:
> Ideally we can establish a pattern that doesn't classify a module as _submodule_ - where all modules using node-config work in both modes.

I'm not sure this is desirable. In some cases you would like to have a completely separate and independent config instance, like when using 3rd party packages. A package owner in most cases shouldn't pollute the configuration instance of its users. For these cases I differentiated between `Config.create()` and `Config.subModule()`. The former creates a completely new instance while the later creates an instance derived from the original configuration instance, and keeps the connection to it. I believe this is an important distinction.

Can you provide some sort of "code samples" of how you imagine the functionality to look and function like?

lorenwest:
First some definitions to clarify communication

* **module** - Any module that has a package.json file and a ./config directory.
[...]

PR #670, opened 2022-02-07
Improved submodule support

### LINKED ISSUE #226: Find a solution to the multiple-instance problem for submodule defaults

rayvincent2:
# Goals
* Ensure modules that leverage node-config could be loaded dynamically (well after `config.get` was called) and the applications configuration became immutable (Issue #576 - Late Loading)
* Ensure that a submodule could have multiple instantiated classes which could leverage the modules default/overridden configs (Issue #226)

# Implications (Backwards Compatibility)
* `setModuleDefaults` will now throw an error if it's called more than once for the same `moduleName`.
  * To ensure consistency for each config instance used in the module, we must enforce that the module only register it's defaults once.  Likely in the module's main file.  I don't expect this to be an issue, but would need your feedback.  I don't see how calling this method more than up to now would give predictable behavior.
* `getModuleConfig(moduleName: string, options: object | undefined)` method was added to return a new `Config` instance which merges the provided options into the module's registered default configs and returns a new Config instance.  This method will throw an error if the `setModuleDefaults` was never called for the given module.
* The Wiki page will need some updates to the submodule page to showcase a few approaches.
* Everything else is 100% backwards compatible.  See the added unit tests for new features.  All unit tests are passing.

# Solved Issues
I started this branch to solve the Late Loading portion of #576, but am of the opinion that all acceptance criteria of #576 are met and the ticket can be closed.  I saw an opportunity to improve module configuration support which closes #226 as well.  Let me know if you'd rather me split this into two PRs.

## Issue #572, Late Loading
I found that the setModuleDefaults method would merge the module defaults into the root config object.  This is obviously a problem when my module is dynamically loaded after `config.get` was called and the root config object was immutable and wouldn't accept my module defaults.

1. I decided to create a separate `moduleConfigs` object to store module default configs.  The module defaults will get any existing overrides in the  global config merged in.  Once set on the global `moduleConfigs` object, only that modules entry will be made immutable immediately.
   * By only setting that current module entry as immutable, it allows for additional modules to be added at any time.
2. The `config.get` method was updated to look in the `moduleConfigs` object for keys if they come back undefined on the global config object.

## Allow multiple instances of modules (issue #226)
Please raise issue with any misunderstanding that I have about the current problems regarding multiple instances of modules.  I read through issues #226 and I believe that I've found that adding one new method `getModuleConfig` provides a clean way to create new Config instances which can share the lifecycle of the module instances.

## Issue #572, Submodule config as default
With the `loadFileConfigs` method, It appears that this is already supported.  As a user of this library, I've never found it difficult to simply add the following lines to my submodule's entry point:
```javascript
var path = require('path');
const baseConfig = config.util.loadFileConfigs(path.join(__dirname, '..', 'config'));
```
This is hardly a burden and the APIs clearly supports this already.

## Issue #572, Dual purpose
Again, if developers write their modules to load configs from the expected file locations, I don't see why this isn't already supported.  If the ask is to allow modules to access config keys without specifying the root module name, then the approach I provided in the examples may help.  Again, I don't see how this isn't already supported.

# Side effects
## Accessing module default configs
Since the default module settings exist separate from the global config object, then you can access the root module configs from any Config instance.  I see this as a unexpected benefit.  This means that the `getModuleConfig` instances created will allow you to access the module's default configurations.
```javascript
config.util.setModuleDefaults('TestModule', { param1: 'value1' });
var moduleConfig = config.util.getModule('TestModule', { param1: "Different Value" });

console.log(moduleConfig.get('param1'));
// Outputs overridden module config "Different Value"

console.log(moduleConfig.get('TestModule.param1'));
// Outputs module default config "value1"
```

# Examples
## Module returns one class
Here's an example module that would provide this behavior.  I added something similar to the unit tests in this PR.
```javascript

// Set the module defaults
config.util.setModuleDefaults('TestModule', {
  test: true,
  example: true
});

function TestModule(options) {
  // Get a unique module config based on your defaults with optional override
  this.moduleConfig = config.util.getModuleConfig('TestModule', options);
}

TestModule.prototype.isExample = function() {
  return this.moduleConfig.get('example');
};

module.exports = TestModule;
```

This allows the application to then use the module multiple times.  Each instance of the TestModule class being able to have unique configuration and override as needed.  The trick simply being that each instance would need to manage the retrieved configuration from `getModuleConfig`.

##  Module returns multiple classes with nested
```javascript

// Set the module defaults
config.util.setModuleDefaults('TestModule', {
  server: {
    host: 'localhost',
    port: 8080
  },
  client: {
    host: 'localhost',
    port: 8080,
    requestTimeout: 10000
  }
});

function Server(options) {
  // Request module configs, but only override server settings
  var moduleConfig = config.util.getModuleConfig('TestModule', {  server: options || {} });
  this.config = moduleConfig.get('server');
}

function Client(options) {
  // Request module configs, but only override client settings
  var moduleConfig = config.util.getModuleConfig('TestModule', {  client: options || {} });
  this.config = moduleConfig.get('client');
}

module.exports = {
  Server: Server,
  Client: Client
};
```

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -19,7 +19,8 @@ var DEFAULT_CLONE_DEPTH = 20,
     env = {},
     configSources = [],          // Configuration sources - array of {name, original, parsed}
     checkMutability = true,      // Check for mutability/immutability on first get
-    gitCryptTestRegex = /^.GITCRYPT/; // regular expression to test for gitcrypt files.
+    gitCryptTestRegex = /^.GITCRYPT/, // regular expression to test for gitcrypt files.
+    moduleConfigs = {}; // Storage for submodule configs
 
 /**
  * <p>Application Configurations</p>
@@ -172,7 +173,17 @@ Config.prototype.get = function(property) {
     checkMutability = false;
   }
   var t = this,
-      value = getImpl(t, property);
+      value = getImpl(t, property),
+      tmpValue;
+
+  // If value is an object it may be only partial, check to see if the 
+  // full object is in dynamicallyImportedModules
+  if (value === undefined || util.isObject(value)) {
+    tmpValue = getImpl(moduleConfigs, property);
+    if (tmpValue !== undefined) {
+      value = tmpValue;
+    }
+  }
 
   // Produce an exception if the property doesn't exist
   if (value === undefined) {
@@ -203,7 +214,13 @@ Config.prototype.has = function(property) {
     return false;
   }
   var t = this;
-  return (getImpl(t, property) !== undefined);
+  var value = getImpl(t, property);
+
+  // If the value is undefined, check to see if it was dynamically loaded later
+  if (value === undefined) {
+    return (getImpl(moduleConfigs, property) !== undefined);
+  }
+  return (value !== undefined);
 };
 
 /**
@@ -226,7 +243,6 @@ Config.prototype.has = function(property) {
  *   });
  * <br>
  *   // Template name may be overridden by application config files
- *   console.log("Template: " + CONFIG.MyModule.templateName);
  * </pre>
  *
  * <p>
@@ -240,6 +256,12 @@ Config.prototype.has = function(property) {
  * @return moduleConfig {object} - The module level configuration object.
  */
 util.setModuleDefaults = function (moduleName, defaultProperties) {
+ 
+  // Throw error if submodule defaults has already been registered
+  var localModuleConfig = getImpl(moduleConfigs, moduleName);
+  if (localModuleConfig !== undefined) {
+    throw new Error('Submodule "' + moduleName + '" has already been set');
+  }
 
   // Copy the properties into a new object
   var t = this,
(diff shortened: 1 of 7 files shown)


markstos (commented):
This appears to solve some some long-standing and difficult problems. I like that it's a backwards compatible solution.

I like to get @lorenwest take on it, who is temporarily less available for personal reasons. I appreciate your patience as we get you more feedback on this. 

Thanks!

lorenwest:
Thank you for this important addition, and the thoroughness of this PR.

I won't have a chance to read through the code, but I agree with the thoughts and concerns addressed.

@markstos will probably give the final thumbs up, but it's a 👍🏽 from my perspective.

markstos:
Thanks for the input @lorenwest ! I want to give this a second read-through before merging.

rayvincent2:
Hi @markstos, I hope all is well.  Do let me know if there's anything that I can do to help you on this.  Whether it's a session to walk through what I've put together or to help with the updated Wiki.  Is it as simple as submitting a change [here](https://github.com/node-config/node-config/wiki/Sub-Module-Configuration/_edit) and then you are pinged to accept the change?

jdmarshall:
> setModuleDefaults will now throw an error if it's called more than once for the same moduleName. 

To an extent I'm looking at the same issue now and I'm not sure what the right answer is. It is certainly possible to end up with multiple copies of the same module in your project. And in general node-config has a last writer wins strategy, so having setModuleDefaults fail on the second call seems to be at odds with that pattern.

Meanwhile, you probably have 2 copies of a module because they're different major/minor version numbers, and one of the reasons you might increment these numbers is due to a change in the configuration. 
Any module that reads back its own configuration would be in trouble in this case. 

Unfortunately I don't have a better suggestion at this time.

rayvincent2:
Very good.  Let's explode this issue and see if I can make sense of it.

For example:
```
application
├── config
│   └── default.json
└── node_modules
    ├── config
    ├── lib1 (1.1)
    │   └── config
    │       └── default.json
    └── lib2
        └── node_modules
            └── lib1 (1.0)
                └── config
                    └── default.json
```

Let's say we have `lib1` which is evolving and has configuration that's evolving too.  We'll naturally get applications and other libraries which will reference different versions of them over time.  In the scenario that I've drawn above, `application` will have config files that's catering to `lib1@1.1`, but it has a dependency `lib2` which is pulling in `lib1@1.0`.

There are two clear problems:
1. If we do throw an error for subsequent calls to `setModuleDefaults` for the same module, then there's no way to cater to this natural evolution of `lib1`.  Which I do agree is a bit of a failure.
2. `application`'s config/default.yaml will override configs for both `lib1@1.0` and `lib1@1.1`.  How can it target which versioned config to override... or should it be able to?  I don't see an existing solution for how to do that today.  It feels related to #226.

I have an idea that may help both.  Let me know what parts of it make sense and what doesn't.
1. I update this PR so that setModuleDefaults/getModuleConfig will still honor the last writer wins strategy for the signature of:
   ```javascript
   setModuleDefaults(moduleName: string, defaultProperties: object): object
   getModuleConfig(moduleName: string, overrideProperties? object): Config
   ```  
    This way you can at the very least allow for a module to evolve and version itself and at the worst, it will require the module developer to artfully craft their configurations to always be future compatible.  Then we still have parity with the current implementation and haven't broken backwards compatibility.

2. Attempt to find a way to namespace registered module defaults by version and simultaneously find a configuration syntax that allows you to target historic versions of the module.  Please beat this up.  If it doesn't make sense yet, I'm hoping that it has legs and can evolve into a solution.
   ```javascript
   /**
    * If version is provided then the Config will store a separate default for each version specified.
    * This will allow module defaults to be set for each version registered by the module owner 
    * (I would only recommend using major.minor as version assuming that the config doesn't 
    * change for bugfixes and to reduce the number of fragmented configurations)
    */
   setModuleDefaults(moduleName: string, defaultProperties: object, version? string): object

   /**
    * This would only be used by the submodule that registered it's defaults and will safely always 
    * return the configuration that it registered overlayed by the `application` configs
    */
   getModuleConfig(moduleName: string, overrideProperties? object, version? string): Config
   ```
   In an attempt at how to update the config syntax to target versions, my brain starts with something like this:
   `application/config/default.json`
   ```json
   {
     "lib1": {
       "theBestFeature": true,
       "logLevel": "info"
     },
     "lib1:1.0": {
       "anOldFeature": false
     }
   }
   ```
   All parser types will essentially convert to json so they can all represent this format in some fashion.  the getModuleConfig can take the default submodule config and then overlay the versioned config on top.  So if `lib1@1.0` and `lib1@1.1` both used the getModuleConfig with version specified they would both respectfully get these configuration overrides set from `application`:
   `lib1@1.0`
   ```json:
   {
     "theBestFeature": true,
     "anOldFeature": false,
     "logLevel": "info"
   }
   ```
   `lib1@1.1`
   ```json
   {
     "theBestFeature": true,
     "logLevel": "info"
   }
   ```
   If we wanted to, we could also throw warnings if config specifies a version that no longer exists so the application has a chance to clean them up.  That may be overkill, but could be done and provided a env var to suppress the warning if needed.  The more I think through this, it's likely better to continue on #226 and not be solved in this ticket.

I know that was a lot... let me know if any of that makes sense.

hershmire:
Is this something that's still getting looked at for consideration?

jdmarshall:
I think that breakdown is interesting but in the end all of this is complicated by node-config being a singleton in a world that knows why singletons are a bad idea.

The “right” solution here is a 4.0 version that supports a constructor call (I have code somewhere that sniffs whether new is being called, I can find it again if anyone wants to know the trick), and then things like behavior by caller can be implemented as arguments. Because at the end of the day it’s not getModuleConfig that needs to behave differently, it’s get(), and adding more arguments for each caller is a race you cannot win.

if you accept that config is not a collaboration between peers creating a “stone soup” of configuration open to anyone to read, and instead is a collaboration between a module and its client, then the config is ten different answers to ten different modules, rather than all things to all modules. You can sort that relationship out at constructor time, rather than at call time or via action at a distance as is currently the case.

jdmarshall:
I suspect the right solution then is to modify first reader wins such that module defaults (configSources?) are part of an instance of node-config, but NODE_CONFIG_DIR gets loaded once, on the first call or on load. Those values get merged in the constructor and marked as read only per instance, instead of globally.

jdmarshall:
Having fiddled with this code again recently, I still like parts of this, but that single writer bit still niggles.

I think now:

if you are trying to only read back your own config, you might be better off using lodash’s defaults. 

If you’re trying to read back someone else’s defaults, such as with defer() (or a coworker who implemented service discovery with a registry intended to be used for string interpolation by dependencies), then you *must* guarantee single writer and then @rayvincent2 ’s solution is correct. 

I’m willing to accept that as a documentation problem and a minor version number change, since several people have reported duplicate defaults calls are broken. I may have fixed that in #816 or it may have been fixed much earlier. That PR certainly will make it easier to use setModuleDefaults though. 

I had forgotten my suggestion of sniffing for new calls, and I’ve lost access to that code. I’m sure we could look at that again, but I may have been barking up the wrong tree since distinct instances could be quite confusing.

rayvincent2:
Sorry I have been radio silent. I have a bit of catching up to do on this, but I do want to share that I've taken this implementation and have been using it without issues (based on our usage) since I posted this in 2022.

I've also gone so far as to convert it to Typescript and to use jest. I've also made it isomorphic so it works nicely in node and when compiled with webpack  for browsers. Would you be interested to see that development?

jdmarshall:
for 4.{0,1} I'm focusing on making sure the jsdoc is up to snuff.  I'm still getting a lot of warning from some of the weird Config functions, many of which are now deprecated.  But it's worth thinking about.

I also may have a solution to allow `setModuleDefaults()` to be called twice, built on having split the config object into 2, and some brainstorming on improving `defer()`. I'm not ready to close that door yet. But every time I dig deeper into the code or some of the old PRs I get a little more understanding of why people are asking for these changes.

markstos:
@rayvincent2 Yes, please share your fork. We are interested both in TypeScript and in alternate test runners.

PR #676, opened 2022-03-31
supporting multiple subdomains

TeleMediaCC:
Please create a new release with my supporting multiple subdomains patch
Thanx!


lib/config.js
```diff
@@ -615,16 +615,18 @@ util.loadFileConfigs = function(configDir, options) {
...
-      if (hostName !== firstDomain) {
-        baseNames.push(hostName, hostName + '-' + env);
-      }
+    var domain = "";
+    hostName.split(".").forEach((subdomain) => {
+      domain += (domain ? "." : "") + subdomain;
+
+      NODE_ENV.forEach(function (env) {
+        // Backward compatibility
+        baseNames.push(subdomain, subdomain + "-" + env);
```
jdmarshall, line 624: so for a machine named server.foo.bar.com, we used to have server-prod.json added to the list, if hostname != 'server'.

If I recall how this code works, each entry earlier in the list is overridden by entries later on, and if so then this is not queueing them in the correct order.

I would expect files that specialize by domain to load in the following order:
com-prod.json
bar-com-prod.json
foo-bar-com-prod.json
server-foo-bar-com.prod.json

I don't know when server-prod.json should load in that sequence, but in the original code it applies last so that probably should be preserved, and I don't think you have the loop structured properly to insist upon that.

I'm wondering what use case you're covering here. Why would I have two machines with the same name in the same environment with different domains? Pet servers with isolation per customer? I think that's still covered by the ordering I proposed above.

markstos: I'm also wondering about the use-case. Please update the pull request description to explain in what cases this would be used and why the existing features of supporting multiple config files don't already address the case well.

markstos:
Closing due to lack of contributor follow-up.

PR #678, opened 2022-06-11
feat: add SUPPRESS_STRICTNESS_CHECK to stop runStrictnessChecks

kaivean:
I only need use config.util.loadFileConfig to load multi config directories.  But now it always runs strictness checks when node-config is required and there are warnings as following:

image

So add env "SUPPRESS_STRICTNESS_CHECK" to stop running strictness checks

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1453,6 +1453,11 @@ util.toObject = function(config) {
 
 // Run strictness checks on NODE_ENV and NODE_APP_INSTANCE and throw an error if there's a problem.
 util.runStrictnessChecks = function (config) {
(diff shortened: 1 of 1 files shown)


kaivean:
> I only need use config.util.loadFileConfig to load multi config directories. But now it always runs strictness checks when node-config is required and there are warnings as following:
> 
> image
> 
> So add env "SUPPRESS_STRICTNESS_CHECK" to stop running strictness checks

@lorenwest Please review the code ,  appreciate it.

markstos:
@kaivean Please comment on whether #343 would resolve or your issue, or how this PR differs.

PR #682, opened 2022-07-03

kruczyna:
Hi team!
I saw https://github.com/node-config/node-config/issues/671 issue and I tried to add the support. I mainly followed how support for JSON5 was implemented, but I failed to understand some things: 
1. I did add JSONC to a catch block to return a soft dependency but I am not entirely sure if it's ok to close up JSON5 within if block entirely
2. I have added a "Configuration file test" but I have failed to understand if the custom parser should also handle the JSONC files?


markstos:
It it necessary to add `jsonc` as a dependency and parser? We already support `json5`, which parses JSON with comments. It seems sufficient to add `.jsonc` to the list of extensions parsed by the JSON5 parser.

markstos:
I'm closing this due to lack of feedback from the developer who wanted the feature. We can re-open if someone expresses interest in the future.

PR #690, opened 2022-09-06
One Spesific Config File Support

efeArdaYildirim:
```sh
export NODE_CONFIG_DIR="/etc/http_app_conf.json"
```

Now it can parse single file.
I change 660-676 and 742-745 lines. Project has no lint config and my vs code is change some spaces, quotes.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,1404 +1,1476 @@
-// config.js (c) 2010-2022 Loren West and other contributors
-// May be freely distributed under the MIT license.
-// For further details and documentation:
-// http://lorenwest.github.com/node-config
+  // config.js (c) 2010-2022 Loren West and other contributors
+  // May be freely distributed under the MIT license.
+  // For further details and documentation:
+  // http://lorenwest.github.com/node-config
 
-// Dependencies
-var DeferredConfig = require('../defer').DeferredConfig,
-    RawConfig = require('../raw').RawConfig,
-    Parser = require('../parser'),
-    Utils = require('util'),
-    Path = require('path'),
-    FileSystem = require('fs');
+  // Dependencies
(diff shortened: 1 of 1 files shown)


lorenwest:
I'd like to know more about the use case for this.  If all you want to do is load a specific config file, why use `node-config` at all?

I'm not a fan of this PR - it specifies a file in an element named for specifying a directory.  If we choose to support this, I'd like to hear some community feedback on the need, and if this is the right implementation path.  

Also the PR would need to include a test to validate the feature and assure it doesn't regress with future PRs.

efeArdaYildirim:
I want to use node-conf because it can parse many conf file formats and it is more human-readable for reading code after a while. Think you write this everywhere in code 
```js
const {logging:{logPath,maxSize}} = JSON.parse(require('fs').readFileSync(process.env['conf'])) // I can make this is a func but I will limit with JSON
// or 
const {maxSize,logPath} = conf.get('logging')
```
 I want to put my conf file into /etc, but node config tries to read all Linux conf. I can add a regex search for files. If I wrote tests and regex filter are you gonna accept them? It is the first contribution I forget to write tests.

lorenwest:
This seems to be a valid use case - reading a config file from a strange
location that you may not control (like /etc), and having it
deterministically merge with other configs from this package.

NODE_CONFIG_DIR can accept a *list* of configuration directories, and while
the name would clash with this use case, allowing a file to be an entry in
this list could resolve a use case of loading a single file that was
generated outside the normal directory structure.

Adding the file to the NODE_CONFIG_DIR list would allow it to be merged in
a deterministic manner among other configs.

I'm open to further discussion on this topic, and yes, we will accept pull
requests with tests.

-Loren

PR #692, opened 2022-09-19
Support loading transpiled JS config files

Tomas2D:
This pull request aims to finish previously opened pull request #566. Adding support for transpiled JS modules (those with `module.exports = config`) is helpful when you have a `config` directory inside your `src` directory in the `TypeScript` project.

To ensure that the file is transpiled and to preserve backward compatibility, I have added a check for `__esModule` and the `default` property that it is an object.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -1,4 +1,5 @@
 // External libraries are lazy-loaded only if these file types exist.
+const util = require("util");
 var Yaml = null,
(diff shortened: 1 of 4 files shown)


Tomas2D:
ping @markstos

Tomas2D:
@markstos would you mind bumping a new version? Thanks!

Issue #689, opened 2022-08-25
setModuleDefaults behaves differently from loadFileConfigs() when the defaults collide with config utility functions


jdmarshall:
### I'm submitting a ... 
  - [ x ] bug report
 
### What is the current behavior?
If your default.js file contains a key that collides with node-config's extendDeep() helpers, such as 'get', then config figures it out and leaves the value. However when you pass in values via setModuleDefaults, they don't get copied because the function already exists.
 
### What is the expected behavior?

setModuleDefaults() and extendDeep() should have the same semantics, so that loadFileConfigs behaves the same as setting defaults.

### Please tell us about your environment:

 -  node-config version: 3.3.7
 -  node-version: 14.20.0
 
### Other information

(e.g. are you using an environment besides Node.js?)

jdmarshall:
This seems to be caused by util.makeHidden(). You can't assign to an object once you've defined a property by the same name.

lorenwest:
Are you suggesting we add support for [Reserved Keywords](https://github.com/node-config/node-config/wiki/Reserved-Words) by allowing `get` to be a valid config key?

markstos:
> setModuleDefaults() and extendDeep() should have the same semantics, so that loadFileConfigs behaves the same as setting defaults.

I disagree. `node-config` is intentional about making the module immutable, and it's not immutable if we open it back up for changes after it's been made immutable. 

However, I'm leaving this open because the documentation could be clearer on this behavior difference.

jdmarshall:
@markstos You've glossed over the real issue and picked on the commentary.

> If your default.js file contains a key that collides with node-config's extendDeep() helpers, such as 'get', then config figures it out and leaves the value. How

That's a bug.

lorenwest:
@jdmarshall - is this a problem beyond the 3 [Reserved Keywords](https://github.com/node-config/node-config/wiki/Reserved-Words) of `get, has, and util`, or is it only a problem with these reserved words?

While it would be better to have zero reserved keywords, it seems like a pretty major undertaking to remove them.

So, yea, if you try to use one of the reserved keywords, undefined results may occur.  I don't think we've thought of all the problems related to trying to use these keywords, or workarounds.

lorenwest:
If the underlying issue is the inconsistency in the way it fails if you try to use one of the 3 reserved keywords, I'd be happy to have a conversation with someone willing to make this failure more consistent, and submitting a PR.

jdmarshall:
[...]

PR #705, opened 2022-12-02
Fix for collision of setModuleDefaults with get(), has() and friends

jdmarshall:
This PR addresses #689

The problem is that defineProperty() defaults to configurable: false, which should only be false in makeImmutable(). If you set it before that, you can't delete properties or modify the property.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -315,14 +315,16 @@ util.makeHidden = function(object, property, value) {
   // If the new value isn't specified, just mark the property as hidden
   if (typeof value === 'undefined') {
     Object.defineProperty(object, property, {
-      enumerable : false
+      enumerable  : false,
+      configurable: true,
     });
   }
   // Otherwise set the value and mark it as hidden
   else {
     Object.defineProperty(object, property, {
-      value      : value,
-      enumerable : false
+      value       : value,
+      enumerable  : false,
+      configurable: true
     });
   }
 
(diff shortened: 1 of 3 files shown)


jdmarshall:
This PR may also address a couple of the bullet points in #226

KernelDeimos:
Giving this a +1, looks sane to me

markstos:
@jdmarshall In the related issue, you said you had a problem that was sometimes triggered and sometimes not. Why was the issue sometimes triggered and sometimes not? From looking at the patch here, either `configurable` is true or false. Ithe current `configurable` behavior here was the bug, wouldn't it fail consistently, not just sometimes?

jdmarshall:
We have a very complicated use case, 3 apps that use some of our modules, a peculiar dev situation, an armful of CLIs, and several NodeJS versions. Some would work and some wouldn’t. That’s what I meant. 

This PR comes with a test that reproduces the underlying issue.

jdmarshall:
Any progress on this? I’ve had to rename a bunch of config and add validation to disallow “get” as a key in our config files to prevent this happening to us. So many config files.

jdmarshall:
I've been using a workaround for this issue for 13 months now.  @markstos what needs to be done here?  The original code misunderstands how defineProperty works in the absence of explicit settings, resulting in calls that do not accomplish what is desired.

Then later attempts to define module defaults fail silently, which is the behavior when attempting to assign values to a property that is not defined as mutable.

jdmarshall:
https://github.com/node-config/node-config/wiki/Reserved-Words Suggests this should not be landed.

PR #706, opened 2022-12-17

ducaale:
This is another [attempt](https://github.com/node-config/node-config/pull/682) to support `.jsonc` extension but using JSON5 parser as advised in https://github.com/node-config/node-config/issues/671#issuecomment-1039794125

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -796,7 +796,7 @@ util.resolveDeferredConfigs = function (config) {
  * .js = File to run that has a module.exports containing the config object
  * .coffee = File to run that has a module.exports with coffee-script containing the config object
  * .iced = File to run that has a module.exports with iced-coffee-script containing the config object
- * All other supported file types (yaml, toml, json, cson, hjson, json5, properties, xml)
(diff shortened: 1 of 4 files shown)


jdmarshall (approved):
LGTM

Personally I would use json5 but seems harmless enough. Slightly slower loading to scan another pattern, but not much.

erfanium (approved):
Thanks

KernelDeimos:
`.jsonc` looks to be a subset of `json5`. Since `.jsonc` isn't as well specified [1](https://github.com/microsoft/vscode/issues/100688) [2](https://github.com/onury/jsonc/issues/2) it might make sense to simply treat `.jsonc` files as `.json5` files.

markstos:
I was about to hit merge on this, but there is a small extra bit of overhead to start checking every file to see if it might be JSONC... at the same time, we about to start parsing both the `.json` and `.json5` extensions with exactly the same parser, JSON5 that was proposed here. I don't think we really need *three* file extension associated with the same parser.

So I'm closing this. If you want to use "JSON with comments"  please just use either the `.json` or `.json5` file extensions. Both already support comments.

Issue #712, opened 2023-01-02
[BUG] JSON5 Vulnerability - Prototype Pollution in JSON5 via Parse Method


iwt-ttaskov:
**Describe the bug**

json5 v2.2.1 is vulnerable to 'Prototype Pollution via Parse Method' (see below):

+ yarn npm audit --recursive
└─ json5: 2.2.1
   ├─ ID: 1085271
   ├─ Issue: Prototype Pollution in JSON5 via Parse Method
   ├─ URL: https://github.com/advisories/GHSA-9c47-m6qq-7p4h
   ├─ Severity: high
   ├─ Vulnerable Versions: <2.2.2
   ├─ Patched Versions: >=2.2.2
   ├─ Via: config
   └─ Recommendation: Upgrade to version 2.2.2 or later

**Expected behavior**

Please upgrade json5 to version 2.2.2 or later.
Thank you in advance :) 

### Please tell us about your environment:

 -  node-config version: 3.3.8
 -  node-version: 16.16.0

alimxk:
This is the vuln: [CVE-2022-46175](https://nvd.nist.gov/vuln/detail/CVE-2022-46175)

markstos:
This was fixed in 3.3.9.

PR #713, opened 2023-01-03
fix(vulnerability): upgrade json5 version from 2.2.1 to 2.2.2

veekays:
json5 dependency is vulnerable in the current version which is `2.2.1`, this PR contains the fix. The vulnerability is resolved in version  `2.2.2`.

This PR resolves the issue #712

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "json5": "^2.2.2"
+    "json5": "^2.2.3"
   },
   "devDependencies": {
     "@types/node": "^7.0.8",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "json5": "^2.2.1"
+    "json5": "^2.2.2"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


dragan-novakovic:
Maybe it will be easier if the version is 2.2.x ?

jdmarshall:
I don't agree that the PR that was filed in JSON5 to address this CVE actually addresses the problem. Can someone take another peek at their PR? 

https://github.com/json5/json5/pull/296/files

This seems to be asserting the bug, not fixing it:

https://github.com/json5/json5/pull/296/files#diff-b266c4cbb2cefa3caf90e138f935e3cf489c6655e84c1ca9d0425bb0d9af9cc6R51-R55

jdmarshall (commented):
I would like to see a test that asserts this is fixed.  Because reading their commit history, I don't think it is.

mboaventura:
Hi all,
According to the [nist](https://nvd.nist.gov/vuln/detail/CVE-2022-46175) this vuln was solved in the `2.2.2` version.

dragan-novakovic:
it has been 2 weeks, maybe bump it to 2.2.3 now ?

PR #721, opened 2023-03-03
Fix source maps when using ts config files, improve performance loading ts config files

andrzej-woof:
This PR addresses https://github.com/node-config/node-config/issues/530 and https://github.com/node-config/node-config/issues/613

The problem was that newly registered `ts-node` took over transpilation of all files.
Using `ignore` option we can exclude transpiling again anything that is not a typescript file.

First regex `(?:^|/)node_modules/` is just the default that is used by `ts-node` to exclude `node_modules` if `ignore` option is unspecified.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -61,6 +61,7 @@ Parser.tsParser = function(filename, content) {
   if (!require.extensions['.ts']) {
     require(TS_DEP).register({
       lazy: true,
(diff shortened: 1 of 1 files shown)


markstos:
@flodaniel does this TypeScript file look right to you?

markstos:
I'm willing to get to try this. If we get more bug reports after the next release, we'll have to revert.

PR #723, opened 2023-03-06
Do not try to make regexes immutable

### LINKED ISSUE #722: [BUG] RegExp are proxified which breaks them
**Describe the bug**
Due to [this commit](https://github.com/node-config/node-config/commit/405e4082b6b015329d4300911bd6998482953ee7), regexes are now broken.

If your configuration contains a regex, you'll get a proxified regex which seems to be ok but it will crash as soon as you try to use it with an error: `TypeError: RegExp.prototype.source getter called on non-RegExp object`

**Expected behavior**
You can put regexes in your config files (in cjs or ts format) and use it in your code

 -  node-config version: 3.3.9
 -  node-version: 16.19.1
 
### Other information

A workaround is to enable `ALLOW_CONFIG_MUTATIONS` but it's not an acceptable solution

CGNonofr:
fixes #722

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -411,7 +411,7 @@ util.makeImmutable = function(object, property, value) {
       });
     } else {
       // Call recursively if an object.
-      if (util.isObject(value)) {
+      if (util.isObject(value) && !(value instanceof RegExp) && !(value instanceof Date)) {
         // Create a proxy, to capture user updates of configuration options, and throw an exception for awareness, as per:
         // https://github.com/lorenwest/node-config/issues/514
         value = new Proxy(util.makeImmutable(value), {
--- a/test/11-regexp.js
+++ b/test/11-regexp.js
@@ -17,6 +17,11 @@ vows.describe('Tests for regexp').addBatch({
       return CONFIG;
     },
 
+    'A regexp should not be proxified': function() {
+      const regExp = CONFIG.get('SomeMore.regexp1')
+      assert(regExp instanceof RegExp && !require('util').types.isProxy(regExp))
+    },
+
     'A regexp should not be replaced': function() {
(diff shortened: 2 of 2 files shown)


markstos:
Please include a test suite test which fails before the change and passes afterwards.

CGNonofr:
> Please include a test suite test which fails before the change and passes afterwards.

Just did in 9b9be2fceb7ccbf0c13865058dac13f97e40dd9d

markstos:
Thank you. I see the code also affects Date objects as well. Please make sure that case is covered by a test as well.

markstos:
Still waiting on the update for Dates.

markstos:
Review the test suite and choose a good location.

jeremy-daley-kr:
@CGNonofr and @markstos ... I believe this is the error I'm also experiencing, introduced in v3.3.8. Any thoughts how we can get this moving?

markstos:
@jeremy-daley-kr I'll merge this once test coverage for the date case is added.

jeremy-daley-kr:
> @jeremy-daley-kr I'll merge this once test coverage for the date case is added.

Ok. Lemme see if I can take over on that.

jeremy-daley-kr:
@markstos Would you mind reviewing my PR to pick up where this one left off. Seemed simple enough to follow the current convention of tests:
https://github.com/node-config/node-config/pull/775

jeremy-daley-kr:
Actually, the more I look at this, the more I wonder if `Config.get()` can just receive an additional param to bypass the mutability check. In some cases, it could mean just editing one line of code (where the `.get()` call exists, rather than changing multiple config files to use `{ key: raw(/some complex regexp/) }`

markstos:
Closing in favor of #775

PR #724, opened 2023-03-08
Fix sourcemap issue when using .ts files

EvanTedesco:
This PR addresses https://github.com/node-config/node-config/issues/530

Current Behavior: Config files with a .ts extension breaks sourcemaps for jest resulting in incorrect line reporting for test failures.  

Expected Behavior: The line number for a test failure should be correctly reported.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -58,15 +58,6 @@ Parser.jsParser = function(filename, content) {
 };
 
 Parser.tsParser = function(filename, content) {
-  if (!require.extensions['.ts']) {
-    require(TS_DEP).register({
-      lazy: true,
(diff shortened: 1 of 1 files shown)


markstos:
@flodaniel Could you review this?

markstos:
I'm closing this because it conflicted with #721 and they both can't be right. This file appears to remove how were were parsing and loading TS files completely.  Before I would consider merging something like this, you'd have to explain why we needed the code before and why removing it wouldn't break things for some users. 

Any case, the change would need to be redone because of the conflict, which could happen in a new PR anyway.

PR #730, opened 2023-03-24
Add convenience method tryGet

KernelDeimos:
- added `tryGet`
- factored out common functionality into decorators
  - `decCheckMutability` adds the checkMutability side-effect to a method
  - `decValidateGetArgs` adds exception on undefined property name
  - `get` and `tryGet` are both decorated with these
- added test for `tryGet`

I decided to make `tryGet` throw an error if it receives `undefined` or `null` as the parameter name. I know `has` has differing behavior, but I _want_ tryGet to throw an exception if I give it an invalid input because that would indicate the code using it has a problem that needs to be fixed.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -196,9 +196,6 @@ decValidateGetArgs = function (delegate, myName) {
  * @return value {*} - The property value, or undefined if it does not exist.
  */
 Config.prototype.tryGet = decValidateGetArgs(decCheckMutability(function(property) {
-  if ( property === '' ) {
-    throw new Error('Empty string is not valid when using tryGet');
-  }
   const t = this;
   return getImpl(t, property);
 }), 'tryGet');
--- a/lib/config.js
+++ b/lib/config.js
@@ -150,6 +150,57 @@ const getImpl= function(object, property) {
 };
 
 /**
+ * Decorates a method to trigger mutability check
+ * @param delegate {function} - The method to delegate to
+ * @return result {function} - The composed function
+ */
+const decCheckMutability = function (delegate) {
+  return function () {
+    // Make configurations immutable after first get (unless disabled)
+    if (checkMutability) {
+      if (!util.initParam('ALLOW_CONFIG_MUTATIONS', false)) {
+        util.makeImmutable(config);
+      }
+      checkMutability = false;
+    }
+
+    return delegate.apply(this, arguments);
+  }
+}
+
+/**
+ * Decorator for get methods (get, tryGet) to validate the property argument
+ * @param delegate {function} - The method to delegate to
+ * @param myName {string} - The name of the method being decorated (for exception message)
+ * @return result {function} - The composed function
+ */
+decValidateGetArgs = function (delegate, myName) {
+  return function (property) {
+    if(property === null || property === undefined){
+      throw new Error("Calling config."+myName+" with null or undefined argument");
+    }
+    return delegate.apply(this, arguments);
+  }
+}
+
+/**
+ * <p>Get a configuration value and silent fail</p>
+ *
+ * <p>
+ * This performs the same behaviour as get but does not throw an exception if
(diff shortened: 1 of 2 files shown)


markstos:
I think the combined behavior of `get()` and `has()` are sufficient for nearly everyone.  It's only after about 10 years that someone proposed `tryGet()`, and one request for popular module used over 10 years does not justify adding it.  You are welcome to publish `tryGet()` as an external module that injects this method into `node-config`.

KernelDeimos:
`¯\_(ツ)_/¯`

lorenwest:
Reserved words are very bad.  We used to have about 20 reserved words (words that would clash with config names), and narrowed it down to just 3 - ```get, has, and util```.

Adding ```tryGet``` to this list will very likely clash with someone that is currently using tryGet as a config name.

KernelDeimos:
> Reserved words are very bad. We used to have about 20 reserved words (words that would clash with config names), and narrowed it down to just 3 - `get, has, and util`.
> 
> Adding `tryGet` to this list will very likely clash with someone that is currently using tryGet as a config name.

Why are these "reserved"? Do you mean I can do `config.someProperty` instead of `config.get('someProperty')`?

I would've implemented something like this instead:

```javascript
config.get('someProperty'); // example
config.someProperty; // undefined

config = config.__getProxy__({ throwOnMissing: false });

config.someProperty; // example
```

markstos:
@KernelDeimos If you are referring to using Node proxy objects, those were added to the language about 5 years after `node-config` as released.

lorenwest:
Reserved words are bad because you can't use them in configurations like this:
```
{
  "got": "Global Object Training",
  "get": "Global Execution Timer"
}
```
You aren't be able to do `config.get('get')` because `get` is a reserved word.  I didn't look at the implementation of `tryGet()` but if it's anything like `get()` then it would make `tryGet` another reserved word.

https://github.com/node-config/node-config/wiki/Reserved-Words

KernelDeimos:
> If you are referring to using Node proxy objects,

not necessarily, the hypothetical `__getProxy__` method could create an object loaded with all the config properties without using Proxy.

> those were added to the language about 5 years after `node-config` as released.

IMO it's a good idea to implement something like this and mark the use of properties populated on the `config` object as deprecated. This is the first configuration utility I've seen that has reserved words and it would make it difficult to migrate or interoperate between other configuration-related modules.

lorenwest:
The two styles of obtaining configurations are
```
var dbHost = config.get('dbConfig.host');
```
and 
```
var dbHost = config.dbConfig.host;
```

And while I agree with your opinion that deprecating these last 3 keywords and dropping support for the second style has some benefit, the communication effort and management through a major version release is huge.

With a few hundred million downloads over the past decade, there are sure to be a few hundred (thousand?) developers that don't agree, and won't appreciate changing their application code in order to keep up with the latest major release.

KernelDeimos:
> the communication effort and management through a major version release is huge

A part of me wants to say "that's what major versions are for", but I realize that's naive and idealistic; it doesn't really work out that way in actuality, does it?

Just because it's an interesting thought though, something like this could be done if the benefit was deemed worthy of the effort:

1. make a copy of `node-config`, with a different name
2. make the change on the copy, so you have to call `config.getProxy()` if you want the 2nd style of obtaining configurations
3. replace `node-config` with a small entry script that depends on the copy, and populates its namespace with all the config variables for backwards-compatibility

lorenwest:
We've been looking for new community leaders to provide continuity over the next decade or so with this project as the current leaders have over the past decade or so.

Let us know if you feel up for making these core architectural decisions and have the energy to manage this project & community over time.

jelveh:
Hey all! 

On behalf of, @HeyPuter, I just wanted to say that we'd love to adopt this repository. (@KernelDeimos is a member of the Puter team)

Please do let us know your thoughts :)

PR #732, opened 2023-04-24
always fallback to json5 when syntax error occurs

### LINKED ISSUE #715: [BUG] Comment stripping for JSON in node 19 doesn't work
**Describe the bug**

With Node 19 (not sure which is the minimum failing version) a json with comments (with .json extension) doesn't work.

**Expected behavior**
I expect that the comments are removed, as described by [the docs](https://github.com/node-config/node-config/wiki/Configuration-Files#file-formats).

**How to reproduce**

<details>
<summary>package.json</summary>

```json
{
  "name": "test",
  "version": "1.0.0",
  "main": "index.js",
  "dependencies": {
    "config": "^3.3.9"
  }
}
```

</details>

<details>
<summary>config/default.json</summary>

```json
{
    // Example comment
}
```

</details>

<details>
<summary>index.js</summary>

```js
const config = require("config");
```

</details>

Running `node index.js` it fails with:

```
/tmp/test/node_modules/config/lib/config.js:853
    ^

Error: Cannot parse config file: '/tmp/test/config/default.json': SyntaxError: Expected property name or '}' in JSON at position 6
    at util.parseFile (/tmp/test/node_modules/config/lib/config.js:853:11)
    at /tmp/test/node_modules/config/lib/config.js:649:26
    at Array.forEach (<anonymous>)
    at util.loadFileConfigs (/tmp/test/node_modules/config/lib/config.js:648:16)
    at new Config (/tmp/test/node_modules/config/lib/config.js:113:27)
    at Object.<anonymous> (/tmp/test/node_modules/config/lib/config.js:1510:31)
    at Module._compile (node:internal/modules/cjs/loader:1239:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1293:10)
    at Module.load (node:internal/modules/cjs/loader:1096:32)
    at Module._load (node:internal/modules/cjs/loader:935:12)

Node.js v19.4.0
```

 -  node-config version: 3.3.9
 -  node-version: 19.4.0
 
### Other information

Probably the message changed (again?) and it fails to fallback to json5

https://github.com/node-config/node-config/blob/4c1c6193903bdb8e477792bec9dd852f7508dbcf/parser.js#L163

Eskibear:
fix #715

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -160,7 +160,7 @@ Parser.jsonParser = function(filename, content) {
   catch (e) {
     // All JS Style comments will begin with /, so all JSON parse errors that
     // encountered a syntax error will complain about this character.
-    if (e.name !== 'SyntaxError' || e.message.indexOf('Unexpected token /') !== 0) {
+    if (e.name !== 'SyntaxError') {
       throw e;
     }
     if (!JSON5) {


markstos:
This only fixes the issue if the error `message` property would always match when it was already of type `SyntaxError`. Is that the case? Can this expressed in a new test in the test suite?

Eskibear:
https://github.com/node-config/node-config/issues/715#issuecomment-1519517416 can be a new test case. Feel free to edit.

BTW, personally I think it can always fallback to use json5 when error occurs. I don't see big problem/overhead giving json5 parser a try. But I'm not familiar with the context you are already doing so, and that's why I made minimum change in this PR.

lorenwest:
If we fallback to JSON5, we lose JSON validation for JSON files.  Not that it's important to me, but it may be important to someone.  Otherwise, why not replace our JSON parsing with JSON5, and always parse with JSON5?

The fact that we're losing something that someone may feel is valuable suggests this would be done on a major version change.

Eskibear:
> Otherwise, why not replace our JSON parsing with JSON5, and always parse with JSON5?

That's the point. Or probably from the beginning it's not an ideal idea to fallback to use JSON5 parser. What I myself expect is, node-config suggest a default parser based on file extension, .json/.json5, and doesn't auto fallback to use other parsers **unless user explicitly specifies to use another one**. 

> The fact that we're losing something that someone may feel is valuable suggests this would be done on a major version change.

Exactly. Given the fact that it already fallbacks to use JSON5 parser for some cases where syntax error occurs, that's why we see issues like #715 now.

markstos:
Closing. See https://github.com/node-config/node-config/issues/715#issuecomment-1664628378

Someone should submit a PR to always use JSON5 parser to parse all JSON files as an alternate solution. 

Thanks.

PR #734, opened 2023-05-12
Introduce check-environment-variables.js

VictorRos:
## Description

Check environment variables defined in `custom-environment-variables.*`

Enabled if `NODE_CONFIG_CHECK_ENV_VARS` is set to `true`.

For each defined environment variables,

1. Check if it exists and use its value
2. Otherwise, fallback to default configuration for the value
3. Not exists, and no fallback, then add environment variable to missing environment variables array

If at least one environment variable is missing, then raise an error with missing environment variables list.

Example of error:

`Error: Missing environment variable(s): SERVICE_NAME`

## History

After using `node-config` in multiple services in my company for multiple months, we quickly realised a missing feature with `custom-environment-variables.js`.

When we deploy our services to production and environment variables were missing, we were not aware at startup that a value could be missing.

So we have created this file named `check-environment-variables.js` to exit early.

In the first version, all defined environment variables in `custom-environment-variables.js` were mandatory.
An error was triggered if at least one environment variable was missing.

Then, we made an enhancement to have them not considered mandatory if we have a fallback value in `default.*` files.

Current workflow is described above and can be simplified like this:

`Check env var --> Fallback --> Add env var to missing array --> Raise error if missing array length > 0`

## Sharing

As we are now using it since a long time, I think it could be interesting to share it with everyone.

Moreover, in our case it could become a native feature.

And we won't have to keep and copy paste in all our services `check-environment-variables.js`

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -4,6 +4,7 @@
 // http://lorenwest.github.com/node-config
 
 // Dependencies
+const checkEnvVariables = require('./check-environment-variables');
 const DeferredConfig = require('../defer').DeferredConfig;
 const RawConfig = require('../raw').RawConfig;
 let Parser = require('../parser');
@@ -18,6 +19,7 @@ let NODE_ENV;
 let APP_INSTANCE;
 let CONFIG_SKIP_GITCRYPT;
 let NODE_ENV_VAR_NAME;
+let NODE_CONFIG_DIR;
 let NODE_CONFIG_PARSER;
 const env = {};
 const configSources = [];          // Configuration sources - array of {name, original, parsed}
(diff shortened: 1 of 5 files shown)


lib/config.js
```diff
@@ -1101,8 +1104,9 @@ util.substituteDeep = function (substitutionMap, variables) {
       }
       else if (util.isObject(value)) { // work on the subtree, giving it a clone of the pathTo
         if ('__name' in value && '__format' in value && typeof vars[value.__name] !== 'undefined' && vars[value.__name] !== '') {
+          let parsedValue;
```
VictorRos, line 1107: Note:
When creating unit tests, I had an error that `parsedValue` was not defined when use line `1114`.

markstos:
I'm in favor of the idea, but not in favor of adding it to node-config, because the code can be completely independent, and node-config is already stretched thin with volunteers to maintain code which can only live in the core library.

You could load a module just before node-config and checks your environment variables.

Your code could be useful to /any/ project to wants to make sure certain environment variables are defined before proceeding.

<checks>

Actually, there are several modules already published that take care of this:

  * verify-env-vars
  * check-env-config
  * jd-getenv

https://www.npmjs.com/search?q=check%20env%20vars

We could link to a module like this from the wiki as a related utility. 

Because this appears to be an independent and already solved problem, I'm going to close this PR.

PR #744, opened 2023-11-03
BunJS: Fix issue with trying to load file that does not exist.

undecidedapollo:
On startup config attempts to load various files, such as "runtime.json". When loading, BunJS returns a different error code than NodeJS for files that do not exist. This updates the check to not throw an exception for this error code.

Let me know if you need any additional information.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -828,7 +828,7 @@ util.parseFile = function(fullFilename, options) {
     fileContent = fileContent.replace(/^\uFEFF/, '');
   }
   catch (e2) {
-    if (e2.code !== 'ENOENT') {
+    if (e2.code !== 'ENOENT' && e2.code !== 'EBADF') {
       throw new Error('Config file ' + fullFilename + ' cannot be read. Error code is: '+e2.code
                         +'. Error message is: '+e2.message);
     }


jdmarshall:
This looks okay to me, but maybe “we” should tap the brakes here.

Are we going to patch ten thousand node packages so work with Bun or are we going to tell Bun to knock it off?

Bun should probably be running unit and integration tests for a corpus of modules as its integration tests. ENOENT seems like a bug to me.

jdmarshall:
I wonder how we go about broaching that subject. ENOENT seems pretty straightforward, and very broad impact.

Jayq69:
You knoe where I stay pull up On Dec 21, 2023, at 11:47 AM, Jason Marshall ***@***.***> wrote:﻿
I wonder how we go about broaching that subject. ENOENT seems pretty straightforward, and very broad impact.

Jayq69:
You know where I stay texaslove On Dec 21, 2023, at 11:47 AM, Jason Marshall ***@***.***> wrote:﻿
I wonder how we go about broaching that subject. ENOENT seems pretty straightforward, and very broad impact.

Issue #704, opened 2022-11-06
[BUG] Failing test on master?


jdmarshall:
**Describe the bug**

`  Configuration file Tests
    ✓ Objects wrapped with raw should be unmodified
    ✓ Inner configuration objects wrapped with raw should be unmodified
    ✓ Supports multiple levels of nesting
    ✗ Supports keeping promises raw by default 
        » expected 'this is a promise result', 
  	got	 undefined (==) // /Users/jasonmarshall/Projects/cobbler/node-config/node_modules/vows/lib/assert/macros.js:14
`

I thought this was something on my end, but I pulled master and I'm still getting this, node 8 through 16

Submitting a PR for a failing test for the test suite is ideal. 

**Expected behavior**
Green tests

**Screenshots**

### Please tell us about your environment:

 -  node-config version: 3.3.8
 -  node-version: 8-18

lorenwest:
Thank you for bringing this to our attention. Not sure where this regression was introduced.

@jdmarshall made their first contribution in https://github.com/node-config/node-config/pull/667
@inside made their first contribution in https://github.com/node-config/node-config/pull/661
@DigitalGreyHat made their first contribution in https://github.com/node-config/node-config/pull/677

Callout to the authors of the last few pull requests - can you verify if your PR introduced this?

inside:
Hi, on master I did:

```
git bisect start
git bisect bad
git bisect good cfde385c024789ae34d0dd731869f0cc59cd9c4c
```

Tests are green on the aforementioned commit. I ran `npm test` for each commit the git bisect command proposed. In the end, git bisect revealed:

```
405e4082b6b015329d4300911bd6998482953ee7 is the first bad commit
commit 405e4082b6b015329d4300911bd6998482953ee7
Author: Grosan Flaviu Gheorghe <fgheorghe@grosan.co.uk>
Date:   Sat Dec 1 11:10:18 2018 +0100

    Fixed property mutation. Throw an exception on such an attempt. Updated tests to capture this exception.

 lib/config.js         | 23 ++++++++++++++++-------
 test/2-config-test.js | 10 +++++++---
 2 files changed, 23 insertions(+), 10 deletions(-)
 ```

I haven't inspected what's in this commit. @fgheorghe can you tell if this is the needle we are looking for?

markstos:
I'm not sure.

markstos:
[...]

PR #745, opened 2023-11-10
fix: lack of comments removal, invalid regexp

DeutscherDude:
This addresses issues mentioned in #715 
The issue mentioned in #704 will be investigated further

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1092,6 +1092,7 @@ util.substituteDeep = function (substitutionMap, variables) {
   const result = {};
 
   function _substituteVars(map, vars, pathTo) {
+    let parsedValue;
     for (const prop in map) {
       const value = map[prop];
       if (typeof(value) === 'string') { // We found a leaf variable name
@@ -1102,7 +1103,7 @@ util.substituteDeep = function (substitutionMap, variables) {
       else if (util.isObject(value)) { // work on the subtree, giving it a clone of the pathTo
         if ('__name' in value && '__format' in value && typeof vars[value.__name] !== 'undefined' && vars[value.__name] !== '') {
           try {
-            const parsedValue = util.parseString(vars[value.__name], value.__format);
+            parsedValue = util.parseString(vars[value.__name], value.__format);
           } catch(err) {
             err.message = '__format parser error in ' + value.__name + ': ' + err.message;
             throw err;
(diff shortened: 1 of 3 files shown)


DeutscherDude:
Hello @lorenwest,
I have amended the parsing of JSON to utilize JSON5 parser as per (this thread)[https://github.com/node-config/node-config/issues/715#issuecomment-1814967711].
Additionally, I removed the extraneous stripping of comments from CSON parser.
There are *some* changes introduced to stripComments, i.e. a new RegExp, which I amended, so that it does not use the [a-zA-z] honey pot.
Happy to introduce any further required changes :smile:

timturnerwhcc:
@lorenwest @DeutscherDude There hasn't been much movement on this PR in a bit - it would be really helpful to have as my team is migrating to Node20.

I'm happy to give feedback on this PR or whatever else to help move it forward.

jfberry:
We are also holding off a node-20+ migration because of this bug

lorenwest:
This looks like a healthy change, and I wish it were as simple as approving and merging.  Unfortunately for health reasons I've been less than active in this project, and it seems that 2 test failures have crept into the master branch.

One of the problems was resolved in this branch, but there's another that needs further research before merging.  When did it start failing?  Is it a failure of the test or a failure of the baseline code? Does it only fail on one NodeJS version or many?  What was the original intent, and is it still relevant?  All of this takes time and energy to resolve, and my philosophy is to never merge into master with a single test failure, otherwise the effort everyone has put into these tests are diluted.

@markstos and I are the two authorized to merge into master and push a new release, and we've been doing it for 8-12 years, and neither of us is getting any younger.  Would anyone like to start taking the reigns for node-config going forward?  It's pretty stable, but does need TLC every now and then.

jfberry:
I generally agree with your philosophy but it sounds like this pr doesn’t increase the failures that exist, in fact reduces them?
If so it’s a positive step forward. 
Besides what is a show stopper for people with particular config files upgrading to newer versions of node.

DeutscherDude:
Happy to be of further aid in the future. I got pretty intimate with the code and feel comfortable with it :).
For now, I am rather entangled with finishing my engineering degree, but gimme a month 😊

lorenwest:
Thanks for the offer @DeutscherDude .  For now I've disabled the failing test as I couldn't see how it could ever have worked.  Your branch was merged, a new release was created, and npm latest is now at v2.3.10

Shayan1191:
It's very good

Issue #746, opened 2023-11-15
[BUG] Configuration files with different extensions and APP_INSTANCE not loaded in the right order


sgaestel:
**Describe the bug**
I having config files with different extensions, and some defined using an APP_INSTANCE, the loading order is not the expected one.

Let's say I have following files:
```
development-instance.json
development-instance.yml
development.yml
```

They are currently loaded as:
```
development-instance.json
development.yml
development-instance.yml
```

**Expected behavior**
From the documentation (and checking at the extensions loading order), I would expect files to be loaded in following order:
```
development.yml
development-instance.json
development-instance.yml
```

### Please tell us about your environment:

 -  node-config version: 1.31.0
 -  node-version: 16.20.2

lorenwest:
You're right that the expected and documented behavior is to load in this order

```
development.yml
development-instance.json
development-instance.yml
```

It would be a bug if it loaded development-instance.json before development.yml.

Would you or a community member be willing to validate this bug by writing a failing test?

sgaestel:
Submitted a PR with both the test and the fix [here](https://github.com/node-config/node-config/pull/747)

PR #747, opened 2023-11-16
Fix GH-746 - Loading config files order with APP_INSTANCE

sgaestel:
This will fix the loading order when having different file extensions and working with APP_INSTANCE as well.
Instead of doing a `forEach` on basenames and extensions, then adding the `APP_INSTANCE`, the filename using the `APP_INSTANCE` is defined before going through all extensions, so that it is loaded after all extensions without the APP_INSTANCE being defined.

Did a console.log (that I removed) for the `allowedFiles` object when running the added test to show the difference:
Before the fix:
image

After the fix:
image

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -638,12 +638,15 @@ util.loadFileConfigs = function(configDir, options) {
   let resolutionIndex = 1;
   const extNames = Parser.getFilesOrder();
   baseNames.forEach(function(baseName) {
-    extNames.forEach(function(extName) {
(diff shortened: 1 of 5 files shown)


lorenwest:
This looks really good.  I'm surprised it took this long to discover - probably because mixed file types with instances isn't very common.

Thank you for finding, fixing, and writing the test!  I'll bring it in and so some local testing, then push a new release with this asap.

jdmarshall:
Thanks for landing this.

Issue #761, opened 2024-04-20
[BUG] YAML binary type is mangled


Maia-Everett:
YAML supports the `!!binary` syntax for declaring binary values, which is supported by `js-yaml` and is parsed as `Uint8Array` values, but `config` mangles them by turning them into objects.

For a minimal example, see:

https://codeberg.org/Linneris/config-yaml-binary-bug

Run:

```bash
npm install
npm start
```

**Expected output**

```
Expected secret: Uint8Array(6) [ 0, 1, 2, 3, 4, 5 ]
Actual secret: Uint8Array(6) [ 0, 1, 2, 3, 4, 5 ]
Buffer equality: true
```

**Actual output**

```
Expected secret: Uint8Array(6) [ 0, 1, 2, 3, 4, 5 ]
Actual secret: Uint8Array { '0': 0, '1': 1, '2': 2, '3': 3, '4': 4, '5': 5 }
node:buffer:535
  if (obj.length !== undefined || isAnyArrayBuffer(obj.buffer)) {
          ^

TypeError: Method get TypedArray.prototype.length called on incompatible receiver [object Object]
    at get length (<anonymous>)
    at fromObject (node:buffer:535:11)
    at Function.from (node:buffer:312:15)
    at file:///[dir]/config-yaml-binary-bug/src/index.js:10:29
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:113:12)
```

### Please tell us about your environment:

 -  node-config version: 3.3.11
 -  node-version: 20.11.00
 -  js-yaml version: 4.1.0
 
Edit: One way to fix this without breaking immutability would be to convert typed arrays to regular arrays with `Array.from()`, then freeze them. But since `Buffer` instances are already returned unmodified, perhaps the simplest fix would be to add

markstos:
@jeremy-daley-kr Could you take a look at this? It seems to be another case like Regexp and Date to handle in the same way.

jeremy-daley-kr:
@Maia-Everett Can you give this a try and let me know if it solves your issue please? Thanks!
https://github.com/node-config/node-config/pull/779/files

Hassan222-ux:
> Array.from

Buffer
TypedArray.prototype.length

markstos:
[...]

PR #762, opened 2024-04-20
Fix conversion of Uint8Array to object (#761)

Maia-Everett:
As said in #761, `node-config` handles `TypedArray` values (including `Uint8Array`) incorrectly, wrapping them in proxy objects that are not accepted by anything that expects a real `TypedArray` (e.g. `Buffer.from()`).

I hit this bug with `js-yaml` 4.x, which made a breaking change compared to `js-yaml` 3.x, returning `Uint8Array` for `!!binary` values rather than returning `Buffer` like 3.x did.

This PR:
* Makes `makeImmutable()` ignore `TypedArray` and `DataView` values (leaving them mutable, as there's no way to make them immutable)
* Makes `cloneDeep()` correctly clone `TypedArray` and `DataView` values
* Bumps the `js-yaml` dev dependency to the latest version 4.1.0, which hits the bug in `node-config`
* Adds a test for the bug fixed

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -372,6 +372,12 @@ util.makeImmutable = function(object, property, value) {
   if (Buffer.isBuffer(object)) {
     return object;
   }
+
+  if (util.isTypedArray(object)) {
+    Object.preventExtensions(object);
+    return object;
+  }
+
   let properties = null;
 
   // Backwards compatibility mode where property/value can be specified
@@ -1014,6 +1020,18 @@ util.cloneDeep = function cloneDeep(parent, depth, circular, prototype) {
       child = Buffer.alloc(parent.length);
       parent.copy(child);
       return child;
+    } else if (util.isTypedArray(parent)) {
+      if (parent.slice) {
+        // Note that Buffer's slice() semantics return a view of the original buffer,
+        // but we've checked for Buffer already
+        child = parent.slice();
+      } else if (DataView && parent instanceof DataView) {
+        child = new DataView(parent.buffer.slice());
+      } else {
+        // Unknown kind of typed array - shouldn't happen
+        child = new Uint8Array(parent.buffer.slice());
+      }
+      return child;
     } else {
       if (typeof prototype === 'undefined') child = Object.create(Object.getPrototypeOf(parent));
       else child = Object.create(prototype);
@@ -1340,7 +1358,7 @@ util.extendDeep = function(mergeInto) {
 /**
  * Is the specified argument a regular javascript object?
  *
- * The argument is an object if it's a JS object, but not an array.
+ * The argument is an object if it's a JS object, but not an array or a typed array.
  *
  * @protected
  * @method isObject
@@ -1348,7 +1366,7 @@ util.extendDeep = function(mergeInto) {
  * @return {boolean} TRUE if the arg is an object, FALSE if not
  */
 util.isObject = function(obj) {
-  return (obj !== null) && (typeof obj === 'object') && !(Array.isArray(obj));
+  return (obj !== null) && (typeof obj === 'object') && !(Array.isArray(obj)) && !(util.isTypedArray(obj));
 };
 
 /**
@@ -1361,6 +1379,18 @@ util.isObject = function(obj) {
  */
 util.isPromise = function(obj) {
   return Object.prototype.toString.call(obj) === '[object Promise]';
+};
+
+/**
+ * Is the specified argument a TypedArray?
+ *
+ * @protected
+ * @method isTypedArray
+ * @param obj {*} An argument of any type.
+ * @return {boolean} TRUE if the arg is a TypedArray, FALSE if not
+ */
+util.isTypedArray = function(obj) {
+  return ArrayBuffer.isView(obj);
 };
 
 /**
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "coffeescript": "2.2.4",
     "cson": "^3.0.1",
     "hjson": "^1.2.0",
-    "js-yaml": "^3.2.2",
+    "js-yaml": "^4.1.0",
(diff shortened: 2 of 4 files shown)


markstos (approved):
LGTM.

markstos:
@lorenwest This PR seems consistent with some other special cases we handle and includes test coverage. I recommend merging it.

markstos:
A competing PR was merged, but there are parts I like better here. I like that this returns `Uint8Array` and not a buffer, and I think the test coverage here was more clearly organized. I would like to get the test coverage contributed for both cases merged like this:

```javascript
'use strict';

const requireUncached = require('./_utils/requireUncached');
require('../parser');

'use strict';

const vows = require('vows');
const assert = require('assert');

vows.describe('Tests for parsing binary data')
  .addBatch({
    'Using the YAML parser - Uint8Array': {
      topic: function() {
        process.env.NODE_CONFIG_DIR = __dirname + '/22-binary';
        return requireUncached(__dirname + '/../lib/config');
      },
      'A binary value should be immutable': function(config) {
        assert.throws(() => {
          const auth = config.get('auth');
          auth.secret = new Uint8Array([ 1, 2, 3 ]);
        }, /Can not update runtime configuration/)
      },

      'reading !!binary returns a real Uint8Array': function(CONFIG) {
        assert.deepStrictEqual(CONFIG.get('auth'), {
					secret: new Uint8Array([0x10, 0x11, 0x12, 0x13, 0x14, 0x15])
				});
      },
    }
  })
  .export(module);
```

jdmarshall:
I don’t think exposing mutable data was meant to be part of the Config design and API. 

In fact I’m not happy with the current Buffer behavior referenced at the top of this function. That seems wrong. It should at least be a copy operation. 

I believe if you’ve disabled immutability this code doesn’t get called at all, so we are just talking about the default behavior of disallowing modification after initialization. 

Mutability make the config into global shared state. There’s a place for that, and if you can overlay a single writer api over the top of node-config you can do a lot of cool stuff. As long as every function you call can’t modify that state it’s chaos for local reasoning. Not to mention testing. You might as well just use global.*. 

Global shared state doesn’t scale past a few ambitious coworkers trying to “get things done” and deal with consequences “later”. If you’ve picked up node-config you probably trying to avoid that 

I think I’d be open to snapshotting. And a new TypedArray doesn’t have to be built on a Buffer. It can be created on an Array built from the source buffer. Though how are you even getting a buffer into config in the first place?

markstos:
Ancient versions of the library did support mutability. Making the config structure immutable was something I advocated and Loren agreed, and keeping the config state immutable has been a priority since.

jdmarshall:
Good, then we are all three in accord (I've had this conversation before with Loren as well).

I think there is a hole here, that I would like to make smaller or close if we can.  But I think this solution is going left when we want to go right.

An part of that is knowing more about the use case. I'm still in the 'why would you encounter this?' bubble and that's not a good place to make a counterproposal from.

markstos:
> I'm still in the 'why would you encounter this?' bubble and that's not a good place to make a counterproposal from.

There's a test case to illustrate. It seems it can happen when you store binary objects in a YAML config file.

markstos:
It looks like there's alignment about merging this, but now there are conflicts that need to be resolved.

lib/config.js
```diff
@@ -372,6 +372,12 @@ util.makeImmutable = function(object, property, value) {
   if (Buffer.isBuffer(object)) {
     return object;
   }
+
+  if (util.isTypedArray(object)) {
```
jdmarshall, line 376: I want to be clear that I'm not trying to shoot the messenger here.

This function is a mess. And every case we add to it makes it more obvious to me that it's a mess. 

It's a bad copy of cloneDeep, and I'm encountering places where we are recursing objects we should not be recursing. I think this needs a rethink. I'm trying to write a benchmark loop right now to make sure I'm not introducing regressions, and it's breaking because somehow we've gotten past the Buffer guard at the top of the function and now I'm getting errors from the TypedArray inside the Buffer. So there's that.

I just discovered the other day that the reason `makeImmutable()` couldn't be moved to Util is actually false coupling. So I think it can be moved, leaving the one, intrusive backward compatibility feature in config.util (and deprecating it) and rearrange this mess to look more like cloneDeep, which it should in fact be a sibling to and so very much isn't. Probably because different people have cared about different bugs in each.

I have enough open PRs at the moment, with two more in the chamber, so I won't be tackling that right now but for sure that's on the roadmap now.

lib/config.js
```diff
@@ -1014,6 +1020,18 @@ util.cloneDeep = function cloneDeep(parent, depth, circular, prototype) {
       child = Buffer.alloc(parent.length);
       parent.copy(child);
       return child;
+    } else if (util.isTypedArray(parent)) {
```
jdmarshall: I don't think we need useBuffer anymore. That's ancient versions of Node now. And Buffer could be pulled down into this block to simplify the amount of testing we are doing per node in the graph.

jdmarshall:
There's already a change earmarked for 4.4 that affects how makeImmutable works. Might as well make it two to reduce the number of fire drills users have to experience.

lib/config.js
```diff
@@ -546,53 +957,564 @@ util.attachProtoDeep = function(toObject, depth) {
...
+    } else if (useBuffer && Buffer.isBuffer(parent)) {
+      child = Buffer.alloc(parent.length);
+      parent.copy(child);
+      return child;
+    } else if (util.isTypedArray(parent)) {
+      if (parent.slice) {
+        // Note that Buffer's slice() semantics return a view of the original buffer,
+        // but we've checked for Buffer already
+        child = parent.slice();
+      } else if (DataView && parent instanceof DataView) {
```
jdmarshall: Are we really going to encounter DataView?

Issue #614, opened 2020-07-30
Setting NODE_CONFIG_DIR to multiple paths causes OS level ENAMETOOLONG error


alubchuk:
### I'm submitting a ... 
  - [x] bug report
  - [  ] feature request
  - [  ] support request or question => Please do not submit support request or questions here, see note at the top of this template.

### What is the current behavior?
While using `NODE_CONFIG_DIR` variable with multiple configuration subfolders, which are quite deeply nested in the filesystem, I'm getting `ENAMETOOLONG` error, as follows:
```bash
{ paths:
   [ 'environment-000',
     'environment-001',
     'environment-002',
     'environment-003',
     'environment-04',
     'environment-05',
     'environment-06',
     'environment-07',
     'environment-08',
     'environment-09',
     'environment-10',
     'environment-11' ] }

[...]

PR #766, opened 2024-06-03
Check if runtimeJson exists before attempting to parse

ZachFontenot:
I'm working on a project using this library that runs into `ENAMETOOLONG` on OSX due to the raw prepending of the entire config dir value. `fs.existsSync` simply returns false in this case instead of blowing up.

This closed issue mentions the same problem https://github.com/node-config/node-config/issues/614

In our case, we use configs separated by business domain as well as environments to keep things clear and specific, if there is another alternative solution that you would prefer, I would happily implement it.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -577,9 +577,6 @@ util.loadFileConfigs = function(configDir, options) {
   APP_INSTANCE = util.initParam('NODE_APP_INSTANCE');
   CONFIG_SKIP_GITCRYPT = util.initParam('CONFIG_SKIP_GITCRYPT');
 
-  // This is for backward compatibility
-  const runtimeFilename = util.initParam('NODE_CONFIG_RUNTIME_JSON', Path.join(dir , 'runtime.json') );
-
   NODE_CONFIG_PARSER = util.initParam('NODE_CONFIG_PARSER');
   if (NODE_CONFIG_PARSER) {
     try {
@@ -704,7 +701,7 @@ util.loadFileConfigs = function(configDir, options) {
   util.extendDeep(config, customEnvVars);
 
   // Extend the original config with the contents of runtime.json (backwards compatibility)
-  const runtimeJson = util.parseFile(runtimeFilename) || {};
+  const runtimeJson = util.getRuntimeConfig(dir, "json");
   util.extendDeep(config, runtimeJson);
 
   util.resolveDeferredConfigs(config);
@@ -1152,6 +1149,20 @@ util.getCustomEnvVars = function (configDir, extNames) {
   });
   return result;
 };
+
+util.getRuntimeConfig = function (configDir) {
+  let result = {};
+  const locatedFiles = util.locateMatchingFiles(configDir, { 'runtime.json': 10 });
+  if (locatedFiles.length === 1) {
+    const configObj = util.parseFile(locatedFiles[0]);
+    if (configObj) {
+      result = configObj;
+    }
+  } else if (locatedFiles.length > 1) {
+    throw Error(`Found multiple files matching for runtime.json`);
+  }
+  return result;
+}
 
 /**
  * Return true if two objects have equal contents.
--- a/test/2-config-test.js
+++ b/test/2-config-test.js
@@ -471,7 +471,7 @@ vows.describe('Test suite for node-config')
     },
     'Multiple config directories': {
       'Verify first directory loaded': function() {
-        assert.equal(CONFIG.get('Customers.dbName'), 'from_default_xml');
+        assert.equal(CONFIG.get('Customers.dbName'), 'override_from_runtime_json');
       },
       'Verify second directory loaded': function() {
         assert.equal(CONFIG.get('different.dir'), true);


markstos:
I have the same questions as I did in #614 I don't understand you end up loading hundreds of configuration values and putting them into a single path when you only run one app at a time. If you could submit a failing test to the test suite that would be clearer. Note that the resolution in #614 was that the user realized they should be orgazing things more efficiently.

Inside of `parseFile()`, there is already a try/catch around attempting to the read the file. It's not clear why an additional check is being used to avoid entering that try/catch block at all. 

If we were to change something, it would be add some additional check in parseFile()... but it's not clear what to add as it already does try/catch.

ZachFontenot:
> I have the same questions as I did in #614 I don't understand you end up loading hundreds of configuration values and putting them into a single path when you only run one app at a time. If you could submit a failing test to the test suite that would be clearer. Note that the resolution in #614 was that the user realized they should be orgazing things more efficiently.
> 
> Inside of `parseFile()`, there is already a try/catch around attempting to the read the file. It's not clear why an additional check is being used to avoid entering that try/catch block at all.
> 
> If we were to change something, it would be add some additional check in parseFile()... but it's not clear what to add as it already does try/catch.

The issue is that the raw concatenated multi-prefix path of config directories, i.e. `:` or `;` delimited directory names, is being directly passed as a single directory prefix in the runtimeJson check, rather than as in other places, consumed via `.forEach`. We have maybe a dozen specific config files across our app and a colleague of mine experienced this because his app directory was more deeply nested in his filesystem. 

The `try/catch` prevents our app from starting because of this `runtimeJson` function call, because the catch only passes on `ENOENT`. 

I think a nicer solution would be to call `.split(DELIMITER)` and take the head for `runtimeJson` or to check if the fs delimiter is present in the string before passing it to `fs.readFileSync`

DirtNap:
Yeah, I'm the author of the module which wraps `node-config` in our implementation, and I just want to highlight two things from what @ZachFontenot said:

1. The reason we end up with a longer `NODE_CONFIG_ENV` is because our project is structured like:

  /config
  /some-functional-area
  /some-functional-area/config
  /other-functional-area/
  /other-functional-area/config

  and so at import-time our wrapper (among other things) discovers every `config` directory under the root of the package and updates `NODE_CONFIG_ENV` to include all of them (plus, obviously, anything existing in `NODE_CONFIG_ENV`).  The value of this is that developers of, say, `other-functional-area` can independently populate `default.json`, `local.yml`, `prod-default-1.js`, etc., for the config of _their specific functional area_, and developers of `some-functional-area` can independently do the same, while still relying on `node-config` for handling file hierarchy, processing of `custom-environment-variables`, util functions, etc.

2. Even if we did not do that, and we had a very simple `NODE_CONFIG_ENV` like, say, `/app/config:/app/patch/config`, it would never be appropriate to try to find `runtime.json` at `/app/config:/app/patch/config/runtime.json`.  We aren't sure what the legacy behavior this method is trying to support is, but it isn't going to be this, because this will find neither `/app/config/runtime.json` nor `/app/patch/config/runtime.json`.   Unless someone specifically curates a config directory with a `:` in it (which will bork all the rest of the file resolution throughout the module) and places `runtime.json` there, this is never going to resolve and is going to _generally_ return `ENOENT` -- so this PR is specifically because `ENOENT` isn't the only error condition that can result from the (very) underlying call to `fopen`, which can also error with `ENAMETOOLONG` when the path exceeds the system max.  `fs.readFileSync` ultimately calls `fstatat`, which _doesn't_ throw `ENAMETOOLONG` when it is salty about the path size, and is more consistent than relying on the side effect of `ENOENT` being thrown.

As @ZachFontenot mentioned, we're also happy to send a patch for the underlying issue with the path construction, but we're not sure what the legacy behavior was supposed to be, but either resolution would be fine for us.

ZachFontenot:
Alright, I have altered the code to just check against `ENAMETOOLONG` and print a warning about it instead of throwing inside of `parseFile()`. I've also added a branch on my fork with broken tests, but it's pretty ugly and I wouldn't want to add these tests to your codebase, so I'll just link it here. https://github.com/ZachFontenot/node-config/tree/breaks_tests

Let me know if anything else is necessary from my end, I think we've laid out a decent explanation of this and how it affects our application.

lorenwest:
I'm so confused.  @DirtNap talks about multiple values for `NODE_CONFIG_ENV`, when there should only be one value - the running environment.  [The docs](https://github.com/node-config/node-config/wiki/Environment-Variables#node_config_env) call it out as a singleton.  Are you instead talking about multiple values of `NODE_CONFIG_DIR`? 

Now, if @DirtNap really meant `NODE_CONFIG_DIR`, I can see the issue.  Multiple values were implemented poorly - in most cases the implementation failures would fail silently (such as `ENOENT`).  That is, until it fails loudly when the full size of `NODE_CONFIG_DIR` is too big to really be a single directory and it fails with `ENAMETOOLONG`.

This pull request would certainly allow that implementation failure to continue being silent, but it doesn't fix the underlying problem that some features simply don't work if `NODE_CONFIG_DIR` has many values (like processing custom environment variables, runtime.json, maybe others).

@DirtNap - Your wrapper library seems to be a reasonable solution for submodule configuration, but it relies on `NODE_CONFIG_DIR` having multiple values, and that implementation doesn't faithfully process all features 

I think the real solution is to fix the failures and process multiple values of `NODE_CONFIG_DIR` faithfully.  That would be an improvement vs. a propagation of something that shouldn't be happening in the first place.

DirtNap:
@lorenwest right, meant `NODE_CONFIG_DIR`, I have edited the above comment for clarity as well.

ZachFontenot:
I'll match the handling of the general config processing/custom environment vars for the runtime config parsing. I believe that's the only instance where multiple values isn't handled correctly.

ZachFontenot:
I've updated my branch to handle the multiple config directory path. I've made some assumptions about this, mostly that having multiple `runtime.json` files is not something anyone is doing, especially considering that this wouldn't have worked previously anyway. 

I haven't changed it yet, but one test is failing, line 473 in `2-config-test.js`, I assume this is because when multiple dirs are present, the `runtime.json` file was not loading at all, let alone first.

ZachFontenot:
@lorenwest I've updated the value in the test to pass, let me know if the implementation is satisfactory.

ZachFontenot:
Hey, just checking in here. If there's anything more I can do to get this merged, lmk

ZachFontenot:
Hey there, @markstos @lorenwest , pinging again about the viability of this change. We need this on our project, so we plan to fork if we can't get this merged in the next couple of weeks

lib/config.js
```diff
@@ -577,9 +577,6 @@ util.loadFileConfigs = function(configDir, options) {
   APP_INSTANCE = util.initParam('NODE_APP_INSTANCE');
   CONFIG_SKIP_GITCRYPT = util.initParam('CONFIG_SKIP_GITCRYPT');
 
-  // This is for backward compatibility
-  const runtimeFilename = util.initParam('NODE_CONFIG_RUNTIME_JSON', Path.join(dir , 'runtime.json') );
```
markstos, line 581: Removing the use of `initParam()` breaks backcompat because the original code checked in the environment and the CLIl for `NODE_CONFIG_RUNTIME_JSON` while the new code does not.

markstos (changes requested):

lib/config.js
ZachFontenot: ~I can just put this back then I believe?~
I think the problem is the default value using `dir`, which was the initial bug, so what I will do is accept the ENV param location if present and otherwise use my new handling of the runtime.json file.

markstos: It's not quite so simple because the change modifies what's passed to `getRuntimeConfig()`.

`getRunTimeConfig()` now contains similar logic to construct the file name around like 1155.

ZachFontenot:
```javascript
util.getRuntimeConfig = function (configDir, runtimeFilename) {
+  if (runtimeFilename.indexOf(Path.delimiter) == -1) { // default value generated by multiple config dirs
+    return util.parseFile(runtimeFilename) || {};
+  }
  
  let result = {};
  const locatedFiles = util.locateMatchingFiles(configDir, { 'runtime.json': 10 });
  if (locatedFiles.length === 1) {
    const configObj = util.parseFile(locatedFiles[0]);
    if (configObj) {
      result = configObj;
    }
  } else if (locatedFiles.length > 1) {
    throw Error(`Found multiple files matching for runtime.json`);
  }
  return result;
}
```

This is a very simple way of handling it, but it may not be to your liking. This is also assuming the environment is checked first and `runtimeFilename` is constructed as before

markstos: The difference with `initParam` is that it also checks an environment variable first.

ZachFontenot: Right, I'm saying 
```javascript
// This is for backward compatibility
  const runtimeFilename = util.initParam('NODE_CONFIG_RUNTIME_JSON', Path.join(dir , 'runtime.json'));
```
and then passing that filename down to 
```javascript
// Extend the original config with the contents of runtime.json (backwards compatibility)
  const runtimeJson = util.getRuntimeConfig(dir, runtimeFilename);
  util.extendDeep(config, runtimeJson);
...
util.getRuntimeConfig = function (configDir, runtimeFilename) {
  if (runtimeFilename.indexOf(Path.delimiter) == -1) {
    return util.parseFile(runtimeFilename) || {};
  }
  
  let result = {};
  const locatedFiles = util.locateMatchingFiles(configDir, { 'runtime.json': 10 });
  if (locatedFiles.length === 1) {
    const configObj = util.parseFile(locatedFiles[0]);
    if (configObj) {
      result = configObj;
    }
  } else if (locatedFiles.length > 1) {
    throw Error(`Found multiple files matching for runtime.json`);
  }
  return result;
}
```

markstos: I took a deeper look at how runtime.json was used and found the feature was deprecated 10 years ago, back when 1.0 was released.

This variable, NODE_CONFIG_RUNTIME_JSON, has been documented to exist for a long time, nor has `runtime.js`. Neither is mentioned on the wiki.

The idea for this system, pre-1.0, was that you could change some config values at runtime and they would get hot-reloaded, but that hot-reloading feature was removed in 1.0. This bit of code to parse the files was left in for some degree of backwards compatibility, but intended next step here was to finish removing that code.

Why are you using this undocumented `runtime.json` file these days? There are plenty of other config files names you could be using instead.

https://github.com/node-config/node-config/wiki/Configuration-Files

markstos: I think I may understand: You /aren't/ trying to use this feature, but `node-config` is continuing to parse the file anyway, but is not handling the newer feature of separating multiple paths properly, and that's the problem for you. 

So if that understanding is correct, you would also be OK with us finishing ripping out this long-deprecated feature.

ZachFontenot: Yeah sure, sounds fun

ZachFontenot:
In respect to removing the runtime config altogether, I will close this PR and open a new one reflecting the actual intent

PR #772, opened 2024-08-19
Add ability to use custom var for ENV detect

e-ogri:
In our production projects, we need the ability to set our own env variable to determine the current environment.

NODE_ENV bad variant because some libraries can only work with two NODE_ENV values ​​(development and production), but we may have others
NODE_CONFIG_ENV - it seems to us wrong for defining the current environment

We can use our environment variable with NODE_CONFIG_ENV at the same time, but it also seems wrong.

It would be cool if they could pass the name of the variable that will contain the value for the environment

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -545,9 +545,10 @@ util.loadFileConfigs = function(configDir, options) {
   // Initialize
   const t = this;
   const config = {};
+  const customNodeEnv = util.initParam('NODE_CONFIG_ENV_VAR_NAME') || 'NODE_CONFIG_ENV'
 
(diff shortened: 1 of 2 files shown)


jdmarshall:
I would go the other way. Define NODE_CONFIG_ENV off of however you define your environment, in your startup script or docker CMD.  

    #! /usr/bin/env/bash 
    export NODE_CONFIG_ENV = DEPLOY_ENV

jdmarshall:
Please use NODE_CONFIG_ENV for this. Default it off of NODE_ENV in your bootstrapping scripts if you like.

PR #775, opened 2024-12-12
fix: bypass Proxy for RegExp and Date objects

jeremy-daley-kr:
Picks up where https://github.com/node-config/node-config/pull/723 left it, and adds a test for `Date`.

Fixes:
* https://github.com/node-config/node-config/issues/722
* https://github.com/node-config/node-config/issues/694


markstos:
Merged. Are there other issues or PRs that you'd like to see addressed in the next release?

jeremy-daley-kr:
> Merged. Are there other issues or PRs that you'd like to see addressed in the next release?

No, sir. I think I'm good. Speaking of... Is there a cadence or process that new PRs tend to go through before next NPM publish?

markstos:
@jeremy-daley-kr I'd like to get this out soon, but it looks like there's another parallel case with binary files to possibly address first, in #761

jeremy-daley-kr:
> @jeremy-daley-kr I'd like to get this out soon, but it looks like there's another parallel case with binary files to possibly address first, in #761

I'll have to look into this when I get more time to do so. The `Proxy` thing is making a solution more "specific" than I'd prefer with a growing list of `instanceof` checks. If there's another way, I'd like to see if that exists.

I also started exploring adding a `get()` to the `Proxy` config, but don't have something yet that doesn't still require a chain of `|| instanceof`s:

```diff
+ value = new Proxy(obj, {
+   get(target, prop, receiver) {
+     if (target instanceof RegExp || target instanceof Date) {
+       const value = Reflect.get(target, prop, target);
+       if (typeof value === "function") {
+         return value.bind(target); // Bind methods to the target
+       }
+       return value; // Return property value as is
+     }
+     return Reflect.get(target, prop, receiver);
+   },
  set(target, prop, value) {
    throw new Error(`Cannot modify property "${prop}". Object is immutable.`);
  },
});
```

daleyjem:
> Maybe we would better served by declaring the types that we /can/ make immutable rather than the types that we can't.

Can you take a look at this one?
https://github.com/node-config/node-config/pull/779

I feel better about it, because it keeps immutability, and doesn't require specific `instanceof` checks. From what I can tell, it fixes the `binary` issue as well (added test).

Hassan222-ux:
||

Hassan222-ux:
a get()

Hassan222-ux:
Proxyالهدف

Hassan222-ux:
value = new Proxy(obj, {
+   get(target, prop, receiver) {
+     if (target instanceof RegExp || target instanceof Date) {
+       const value = Reflect.get(target, prop, target);
+       if (typeof value === "function") {
+         return value.bind(target); // Bind methods to the target
+       }
+       return value; // Return property value as is
+     }
+     return Reflect.get(target, prop, receiver);
+   },
  set(target, prop, value) {
    throw new Error(`Cannot modify property "${prop}". Object is immutable.`);
  },
});

markstos:
@Hassan222-ux What you are trying to communicate is not clear. Using a translation app to translate what you want to see into English if you need to.

PR #776, opened 2024-12-12
ci: add unit tests to PR CI

jeremy-daley-kr:
I don't think I can see these run in CI without the proper permissions, but seems like it could help get the ball rolling on some kind of PR CI.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,29 @@
+name: PR Checks
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


markstos (approved):
Thanks for this modernization!

jeremy-daley-kr:
@markstos I actually should've been more diligent about confirming it works.
I've added a new PR here to fix something I missed: https://github.com/node-config/node-config/pull/778/files

PR #779, opened 2024-12-18
feat: add `get` trap to Proxy for agnostic prototype access

daleyjem:
## Changes

* Adds `get` trap in `util.makeImmutable`s `Proxy` config
  * Removes `instanceof` checking
  * Fixes https://github.com/node-config/node-config/issues/761
* Add tests to confirm immutability
* Adds a `vows` script so that individual test files can be run
* Adds test to assert serialization with `JSON.stringify`

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -415,15 +415,20 @@ util.makeImmutable = function(object, property, value) {
         // https://github.com/lorenwest/node-config/issues/514
         value = new Proxy(util.makeImmutable(value), {
           get(target, property, receiver) {
-            // Config's own prototype methods (`get`, `has`, etc.)
+            // Config's own defined prototype properties and methods (e.g., `get`, `has`, etc.)
             const ownProps = Object.getOwnPropertyNames(Config.prototype);
-            // Should bypass proxy receiver for properties (e.g. RegExp.prototype.source)
+
+            // Bypass proxy receiver for properties directly on the target (e.g., RegExp.prototype.source)
+            // or properties that are not functions to prevent errors related to internal object methods.
             if (ownProps.includes(property) || (property in target && typeof target[property] !== 'function')) {
               return Reflect.get(target, property);
             }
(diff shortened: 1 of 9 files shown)


markstos:
Looks like an excellent approach with maximum immutability and minimal exceptions. Let's try it.

jeremy-daley-kr:
Sweet! Thanks again, @markstos

markstos:
@jeremy-daley-kr I found there was a competing PR submitted to handle the binary type mangling:

https://github.com/node-config/node-config/pull/762

That PR returns a the binary object as a `Uint8Array`, while your PR returns the binary object as a Buffer that contains Uint8Array's. This was reflected when I tried to merge both and got this test failure about the different types:

  +   secret: Buffer(6) [Uint8Array] [
  -   secret: Uint8Array(6) [

Returning `Uint8Array(6) ` seems simpler and preferable. Could you review why this is returning a buffer instead?

daleyjem:
@markstos I might be a bit confused what the expected behavior is then, because this is the repo that was linked from Maia's original issue (https://github.com/node-config/node-config/issues/761), and it uses `Buffer.from` to assert equality:
https://codeberg.org/Linneris/config-yaml-binary-bug/src/branch/main/src/index.js

daleyjem:
@markstos Also, I just went back to v3.3.7 (before the `Proxy` stuff was added), and added the test snippet you linked to... It gives `Buffer(6) [Uint8Array]`.

daleyjem:
Since that's a new feature request/breaking change, can you consider publishing a new release in the meantime?

jeremy-daley-kr:
@markstos Friendly bump ^^

daleyjem:
> @jdmarshall has been added as a co-maintainer to help move things along. We are evaluating plans for the next release now. Thanks for your patience with our volunteer team.

Thanks, @markstos !

PR #782, opened 2025-01-06
Remove all references to runtime.json and runtime config

ZachFontenot:
@markstos @lorenwest 

I've removed the two calls to generate the runtime config in `lib/config.js` and reverted tests to any value that was present from not loading `runtime.json` files. I've removed the references to `runtime.json` being a valid config file in some of the docstrings as well.

Let me know if there's any tests y'all want filled back in from the absence, which I'll point out via comments


test/2-config-test.js
```diff
@@ -299,7 +299,7 @@ vows.describe('Test suite for node-config')
       assert.equal(CONFIG.get('customerDbPort'), '5999');
     },
     'A cloned property accessor remains a getter': function() {
-      assert.equal(CONFIG.Customers.get('dbString'), 'override_from_runtime_json:5999');
+      assert.equal(CONFIG.Customers.get('dbString'), 'from_default_xml:5999');
```
ZachFontenot, line 302: This is technically "overridden", but maybe not the intended spirit of the test case

test/util.js
```diff
@@ -64,7 +64,6 @@ vows.describe('Tests for config util functions')
         'It can handle recursion': function(util) {
           var result = util.loadFileConfigs(path.join(__dirname, '21-reentrant'));
           assert.ok(result.nested, "did not load nested values");
```
ZachFontenot, line 66: I didn't know what to add or test here, so if anything feels missing, let me know

test/2-config-test.js
markstos: This change looks OK to me.

markstos (approved):
Some history here:

`runtime.json`  and NODE_CONFIG_RUNTIME_JSON were deprecated in `node-config` 1.0, which was released 10 years ago. They have not been documented to exist since then, and should generally not in use.

They existed to support hot-reloading, which was also removed at the same time.

These bits of code were left in at the time for backwards compatibility, but now their mere presence causes a problem because of our support for multiple config paths in a single environment variable.

The call to parse the `runtime.json` was never updated to split up the path into multiple directories, so it can fail with an error like ETOOLONG for some people.

The most straightforward way to resolve that is finish removing this code that was deprecated a decade ago.

To be conservative, we can bump the major version number in case we there are some users are somehow using this ancient, undocumented file.

markstos:
I've approved and nudged Loren to review as well.

ZachFontenot:
@markstos @lorenwest 
Hey folks, just checking in on the status of the v4 build?

PR #786, opened 2025-02-27
Adding support for mjs files

skitscha:
Updating parsers to support mjs configuration files.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -313,7 +313,7 @@ Parser.numberParser = function(filename, content) {
   return Number.isNaN(numberValue) ? undefined : numberValue;
 };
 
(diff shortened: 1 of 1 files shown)


skitscha:
[618](https://github.com/node-config/node-config/issues/618)

skitscha:
@lorenwest Would you be able to take a look at this?

jdmarshall:
Can you add a test for this, in the same style as the other tests for the parser?

jdmarshall:
@markstos is there a reason this couldn’t land in 3.0?

markstos:
It's true that this could go into a 3.0 release. It is not a breaking change.

Issue #787, opened 2025-03-20
[BUG] Don't rely on require.extentions


marco-ippolito:
https://github.com/node-config/node-config/blob/master/parser.js#L70

`require.extensions` has been [deprecated for years](https://nodejs.org/api/modules.html#requireextensions) and its gonna break with Node.js type stripping

lorenwest:
Anyone using the .ts extensions that care to refactor this code?

jdmarshall:
I'm tempted to file a PR against node to un-deprecate this feature.

But then my opinion on this subject is heavily colored by the fact that Domains were deprecated for six years before they got a replacement 2 years ago. I've never seen another programming language deprecate features without a replacement already on-deck. 

I find this confusing and I don't believe that the current Node team has any concrete plans to fix this issue. It's a much, much simpler problem than Domains and it has been festering for longer.

marco-ippolito:
> I'm tempted to file a PR against node to un-deprecate this feature.

Or we can just stop using `require.extensions` since it only works foe commonjs and it conflicts with the internal typescript implementaton

jdmarshall:
This whole thing is in large part about how ts-node figures out if it’s already loaded, and using the same mechanism to see if node-config should load it. 

If you make node-config require nodejs 23 you get built in ts support but until someone is willing to leave a lot of users in the lurch, a different solution is needed. And as I all but said above, I don’t see a good reason to get rid of the extensions mechanism. Especially now that it’s already being copied.

marco-ippolito:
typescript support is enabled by default in Node v23, and the `require.extensions['.ts']` will be true because of that.
Node.js typescript support is different from ts-node so it will break every user. `require.extensions` is an internal and has been deprecated since node v0.x
The reason is that your library will not work AT ALL

Hassan222-ux:
require.resolve.paths(request)

Hassan222-ux:
console.log(require.main);

Hassan222-ux:
> `require.extensionsلقد`

jdmarshall:
Ah. Well, this doesn't play nice with nyc, which I now vaguely recall reading about somewhere else and now I cannot find it again.

jdmarshall:
Okay, I think I now feel the pain of this issue.  Anyone who tries to write integration tests for node-config and combine .ts config files and code coverage from nyc are going to have a bad time, and that's too much loss of test fidelity for me to stomach.

I think I am okay with any of the following solutions:
 - require.resolve()
[...]

PR #788, opened 2025-04-07
Add nyc for code coverage

jdmarshall:
With the prospect of adding additional maintainers, before other people start merging PRs it would be good to have a clearer picture of what we have and that includes code coverage data on the builds. 

nyc@15.1.0 is the most recent version of nyc that is compatible with node >= 10, and I would rather not open engines discussion at this time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "cson": "^3.0.1",
     "hjson": "^1.2.0",
     "js-yaml": "^3.2.2",
+    "nyc": "^15.1.0",
     "properties": "~1.2.1",
     "semver": "5.3.0",
     "toml": "^2.0.6",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -46,8 +47,19 @@
   "engines": {
     "node": ">= 10.0.0"
   },
+  "nyc": {
+    "include": [
+      "*.js",
+      "lib/**/*.js"
+    ],
+    "extension": [
+      ".js"
```
jdmarshall, line 56: This is to keep ts-node and nyc from fighting.

If anyone gets ideas of making node-config into an ESM module we will have to deal with #787 first

markstos:
In terms of supported engines, we support the oldest supported Node LTS release as a policy. That's effectively Node >=20, as Node 18 is being EOL'ed in 3 weeks.

markstos (approved):

jdmarshall:
@markstos i noticed that the comment stripping code had particularly low code coverage. I may file another PR for that.

PR #790, opened 2025-04-09
Improve coverage of parser.js, which is a bit low.

jdmarshall:
Note that the author cited at the top of parser.js now recommends against using his trick, and also has changed it a bit since this was written.

Perhaps we should just switch to using json5 for all .json files and drop this entire function?


markstos:
We already did switch to JSON5 for JSON parsing. Turns out stripComments was dead code since then. See #794

jdmarshall:
This code has been subsequently removed.

PR #791, opened 2025-04-09
Replace arguments fiddling with spread operator.

jdmarshall:
Since none of the code uses more than 3 arguments this doesn't really affect performance much (tested with tinybench) Using arguments deoptimizes JS code, but it's the only spot we're doing this, so it should go.

Spotted while looking at the traversal bug.


markstos (approved):
The new syntax has been well-supported since Node 8, well beyond the cutoff of versions we support.

PR #792, opened 2025-04-10
Fix over-recursion in async and deferred processing

### LINKED ISSUE #610: resolveAsyncConfigs currently "resolves" every character of a string
### I'm submitting a ... 
  - [x] bug report
  - [  ] feature request
  - [  ] support request or question

Currently node-config will attempt to resolve of config objects, using it's build-in resolveAsyncConfigs function. The function recursively iterates over object properties and array elements, in order to resolve them deep. Unfortunately even immutable strings, are currently treated as normal objects, and as strings can be character-wise accessed using [0] to [string.length -1], this behaviour even works. However this causes a *massive* performance drawback when "cloning" large string instances, when a simple "copy" could be enough.

String should not be treated for resolving. So in the resolveAsyncFunction a "typeof prop !== "string"" check should be added.

Also (in case you are just at it), there is i think a small syntax error in the util.extendDeep function, see line 1248:

```javascript
if (mergeFrom[prop] instanceof Date) {
   mergeInto[prop] = mergeFrom[prop];
} if (mergeFrom[prop] instanceof RegExp) {
   mergeInto[prop] = mergeFrom[prop];
} else if ...
```

There is a "else" missing in this line! So I think it would be to change the if into:

```javascript
if ((mergeFrom[prop] instanceof Date) || (mergeFrom[prop] instanceof RegExp)) {
   mergeInto[prop] = mergeFrom[prop];
} else if ...
```

 -  node-config version: latest
 -  node-version: 11
 
### Other information

(e.g. are you using an environment besides Node.js?)

jdmarshall:
for-in will iterate on Strings.

Fixes #610

Adding a sanity check to avoid recursing on them, which should speed up both of these operations substantially.

Also swapping Object.keys() in since we end up filtering by hasOwnProperty() anyway.

### DIFF
--- a/async.js
+++ b/async.js
@@ -41,12 +41,11 @@ function resolveAsyncConfigs(config) {
   var promises = [];
   var resolvers = [];
   (function iterate(prop) {
-    var propsToSort = [];
-    for (var property in prop) {
-      if (Object.hasOwnProperty.call(prop, property) && prop[property] != null) {
-        propsToSort.push(property);
-      }
+    if (prop.constructor === String) {
+      return;
     }
+
+    var propsToSort = Object.keys(prop).filter((property) => prop[property] != null);
     propsToSort.sort().forEach(function(property) {
       if (prop[property].constructor === Object) {
         iterate(prop[property]);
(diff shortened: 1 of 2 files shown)


markstos:
If it passes manual automated testing, LGTM.

PR #794, opened 2025-04-15
drop stripComments - dead code

markstos:
It was no longer in use.

It was removed by 00587019e2013656f470de925d5187929583cbec which switched to using the JSON5
parser for JSON parsing.


lib/config.js
```diff
@@ -1539,8 +1539,7 @@ function _toAbsolutePath (configDir) {
 // Instantiate and export the configuration
 const config = module.exports = new Config();
 
-// copy methods to util for backwards compatibility
-util.stripComments = Parser.stripComments;
+// copy method to util for backwards compatibility
```
jdmarshall, line 1542: Doesn’t this make this a 4.0 change?  Since it’s exposed in util that makes it part of the public interface.

markstos: Yes, I added the v4 milestone, 4.0 label and "breaking-change" labels.

jdmarshall (approved):
LGTM

PR #798, opened 2025-04-16
Fix for iteration bug identified by user @MrFiregore in #731

jdmarshall:
paging @MrFiregore

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -738,19 +738,18 @@ util.loadFileConfigs = function(configDir, options) {
  */
 util.locateMatchingFiles = function(configDirs, allowedFiles) {
   return configDirs.split(Path.delimiter)
+    .filter(dir => dir)
     .reduce(function(files, configDir) {
-      if (configDir) {
-        configDir = _toAbsolutePath(configDir);
-        try {
-          FileSystem.readdirSync(configDir).forEach(function(file) {
-            if (allowedFiles[file]) {
-              files.push([allowedFiles[file], Path.join(configDir, file)]);
-            }
-          });
-        }
-        catch(e) {}
-        return files;
+      configDir = _toAbsolutePath(configDir);
+      try {
+        FileSystem.readdirSync(configDir)
+          .filter(file => allowedFiles[file])
+          .forEach(function(file) {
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -738,19 +738,18 @@ util.loadFileConfigs = function(configDir, options) {
...
-        configDir = _toAbsolutePath(configDir);
-        try {
-          FileSystem.readdirSync(configDir).forEach(function(file) {
-            if (allowedFiles[file]) {
-              files.push([allowedFiles[file], Path.join(configDir, file)]);
-            }
-          });
-        }
-        catch(e) {}
-        return files;
```
jdmarshall, line 752: This returns null if any of the configDir values are falsey.  This is what the stalled PR was trying to fix.

test/20-multiple-config.js
```diff
@@ -56,4 +56,17 @@ vows.describe('Tests for multiple config')
 
   }
 })
+.addBatch({
+  'Empty string should not blow up': function () {
+    process.env.NODE_CONFIG_DIR =  [
+      './test/20-config',
+      ''
```
jdmarshall, line 63: I don't usually like doing this, but sometimes you end up with batch files that end up evaluating a conditional entry to "" and then you're screwed if you're tucking it into NODE_CONFIG_DIR, so I'd rather not have the cryptic error message even if I'd prefel people scrub their inputs. At least the new code should make it a bit more obvious in a step-wise debug of 'where did my configs go?'

lib/config.js
markstos: `.filter(dir => dir)` is tricky to read. I recommend clearer syntax for posterity:

```
.filter(dir => Boolean(dir))
```

This makes the implicit type-casting explicit. Yes, it's unnecessary for the JavaScript interpreter, but it's clearer for humans to read.

markstos (commented):
See also my updated PR description.

jdmarshall:
I think idiomatically people use `filter(Boolean)`. I'm okay with that if you are, but it's not my favorite.

test/20-multiple-config.js
markstos: OK, we can refine later if people complain about the lack of warning message.

PR #800, opened 2025-04-17
Add mjs support

jdmarshall:


jdmarshall:
And now there is a reason this should go into 4.0:  requiring an MJS file requires Node 20 or later in order to function.

risantos:
@jdmarshall could you please update the wiki page?

https://github.com/node-config/node-config/wiki/Configuration-Files#commonjs-javascript-module---js-cjs
> Please note that loading files as [ESM](https://nodejs.org/api/esm.html) is not yet supported. So .mjs and .js when "type":"module" or --experimental-modules will not work.

https://github.com/node-config/node-config/wiki/Special-features-for-JavaScript-configuration-files
> Loading files as [ESM](https://nodejs.org/api/esm.html) is not yet supported. So .mjs or .js when "type":"module" or --experimental-modules will not work.

nikhilmisra63:
i have updated it 4.0.0 but still throwing this error

markstos:
@nikhilmisra63 You have to have a new enough Node version. Check that your Node version supports .mjs files. 

Also, that error could be thrown because your config file can't be parsed because the syntax in the file is wrong. You haven't provided enough information to be certain what the problem is. Can you run `node your-file.mjs` to be certain your Node can parse it at all and that it is free of syntax errors?

nikhilmisra63:
@markstos yes i'm on node 22.11 also all of my files in mjs it just the config module is not working. even i tried to create a sample project where i have just created a default.mjs file inside config folder then a index.mjs with two lines of code

jdmarshall:
What’s the original error? Breakpoint on that line and paste the entire stack trace here.

nikhilmisra63:
[test.zip](https://github.com/user-attachments/files/20359873/test.zip)
here is the sample code.

and this is what i'm getting

/Users/nikhil/Sourcecode/node/dadahub/test/node_modules/config/lib/config.js:862
          ^

Error: Cannot parse config file: '/Users/nikhil/Sourcecode/node/dadahub/test/config/default.mjs': Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/nikhil/Sourcecode/node/dadahub/test/config/default.mjs not supported.
Instead change the require of /Users/nikhil/Sourcecode/node/dadahub/test/config/default.mjs to a dynamic import() which is available in all CommonJS modules.
    at util.parseFile (/Users/nikhil/Sourcecode/node/dadahub/test/node_modules/config/lib/config.js:862:11)
    at /Users/nikhil/Sourcecode/node/dadahub/test/node_modules/config/lib/config.js:667:28
    at Array.forEach (<anonymous>)
    at util.loadFileConfigs (/Users/nikhil/Sourcecode/node/dadahub/test/node_modules/config/lib/config.js:666:16)
    at new Config (/Users/nikhil/Sourcecode/node/dadahub/test/node_modules/config/lib/config.js:115:27)
    at Object.<anonymous> (/Users/nikhil/Sourcecode/node/dadahub/test/node_modules/config/lib/config.js:1524:33)
    at Module._compile (node:internal/modules/cjs/loader:1546:14)
    at Object..js (node:internal/modules/cjs/loader:1689:10)
    at Module.load (node:internal/modules/cjs/loader:1318:32)
    at Function._load (node:internal/modules/cjs/loader:1128:12)

Node.js v22.11.0

jdmarshall:
You've got something else going on and rather than play twenty questions it'd be good if you started with a single, empty mjs file and built up one set of data at a time until you can narrow down the problem with it.

I'm seeing no issues with a simple default.mjs file when I import or require config$4.0.0

```
$ npm ls config
repro@ /Users/jdmarshall/Projects/cobbler/repro
└── config@4.0.0
$ cat config/*

$ cat config/*
export default {
    blah: 3
};

$ cat index.mjs
import config from  "config";

console.log(JSON.stringify(config));

$ npm ./index.mjs
{"blah":3}

$ node --version
v22.14.0

$ nvm use 20
Now using node v20.19.0 (npm v10.8.2)

$ npm ./index.mjs
{"blah":3}

```

nikhilmisra63:
not sure what is real issue but i'm trying to run the same code which i have attached in my last comment and that has only one file and still i'm getting the error.

for now i move to .json instead of .mjs and it is working fine.

jdmarshall:
I tried your sample file. Unable to reproduce. 

> node index.mjs 
> testing config module with mjs

how are *you* starting? You're leaving out bits of your repro case and I haven't left out any of mine.

nikhilmisra63:
Screenshot 2025-05-22 at 1 34 50 PM
here is the complete screenshot please let me know if you need anything else?

jdmarshall:
Anyone else able to repro this case? @markstos @yfp3753 ?

jdmarshall:
Okay. Looks like I need to update the docs and release notes. I misremembered these as being a launch feature and it seems that’s only in 24.

jdmarshall:
Done

PR #802, opened 2025-04-23

rerodrigues:
While **node-config** DO support comments in `.json` and `.json5` file extensions (it uses **json5** parser for both), it doesn't support the `.jsonc` file extension.

Why is this important? Comments are not allowed in the **JSON** specification, and even if they are supported by **node-config**, adding them to a `.json` file will trigger all sorts of errors in IDEs, linters, formatters, code analysis tools, and alike.

Using `.json5` file extension could be a solution, if you're willing to open the doors of your config files to a completely new and full of possibilities file format (**JSON5**). However, if you just want the good ol' **JSON** format, with comments, `.jsonc` is what you need

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -519,11 +519,11 @@ util.getOption = function(options, optionName, defaultValue) {
  * </pre>
  *
  * <p>
- * EXT can be yml, yaml, coffee, iced, json, cson or js signifying the file type.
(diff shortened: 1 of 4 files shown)


markstos (approved):
This being approved because:

- Historically this project has been liberal in the config file formats supported.
- It requires no new parser, so there's very little code.
- JSONC will probably continue to be around for a while a popular config file format.

PR #808, opened 2025-05-16
Stop destructive updates to path in setPath

jdmarshall:
Simplify setModuleDefaults and getImpl/setPath to getPath/setPath

This is prep work for a bigger surgery to split config.js into two pieces.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -128,12 +128,12 @@ const util = Config.prototype.util = {};
  * Underlying get mechanism
  *
  * @private
- * @method getImpl
+ * @method getPath
  * @param object {object} - Object to get the property for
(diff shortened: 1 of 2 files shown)


jdmarshall:
Okay, I have the big PR queued up and it would be a bit smaller if this one landed.

jdmarshall:
Note: this is a little slower, but the slowdown is proportional to the key depth, which for config is usually pretty short.

Local benchmark: shift is about 6 times faster than slice with an array of 100k entries, and twice as fast for an array of 10k entries.

Meanwhile we are more likely to be doing lookups of depth 5-8.

jdmarshall:
I have 2 PRs open at the moment and they're a lot. 

I will file a PR to make this function iterative after those land and that will make this function a lot simpler to trace and remove the slice() calls.

Issue #815, opened 2025-05-17
[BUG] Config.util.extendDeep() breaks on "use strict";


jdmarshall:
Discovered when trying to move util functions to class functions (classes are automatically strict).

This is changing the list of attributes on the object that are getting copied, resulting in exceptions in the tests.

Repro by adding `"use strict";` to the top of lib/config.js and run tests.

jdmarshall:
It blows up in 8-config-extending.js on trying to copy the 'has' function.

I'm not sure it should really be copying the hidden util functions, given that they are added back in with attachProto

>   Actual message: "Cannot assign to read only property 'get' of object '#<Object>'" // /Users/jasonmarshall/Projects/cobbler/node-config/test/8-config-extending.js:31

Issue #806, opened 2025-05-15
[BUG] util.loadFilesConfig() adds entries to configSource()


jdmarshall:
**Describe the bug**
Secondary calls to util.loadFilesConfig() end up adding new configSources in the parseFile() function, which is incorrect behavior, since there is no guarantee the parsed files will end up as part of the configuration. That leads to potential problems when using the information as diagnostic data.

**Expected behavior**
No new entries.

jdmarshall:
The only reason these seem to be getting added is that there's no other way to make this code flow visible, due to the lack of proper separation of initialization and load operations.

Issue #801, opened 2025-04-22
A side effect-free loadConfigs for build time testing and esoteric uses.


jdmarshall:
In #796 I mentioned wanting some of the utils factored out for code cleanliness and standalone use. 

In particular I cited a solution I implemented that made a peculiar use of util.loadConfigs() that I used to support a more robust module defaults.

But there’s another more obvious use and that’s providing validation of the entire config directory. Without forced reloading of node-config which IMO is gross. 

I am proposing a version of loadConfigs that uses no ENV variables and does not modify the internal state like the sources, instead it returns a map that contains all of these modifications, and Config.util.loadConfigs() can call it and then apply the state changes to itself. 

The idea here would be to provide a CLI for node-config that provides a list of NODE_CONFIG_ENV values and can check all files for well-formedness. 

This might be a starting point for evacuating the other incidental util functions that don’t actually utilize the Config object or don’t really need to.

jdmarshall:
For instance, issues like described in #753 would be something that a sanity check would uncover.

markstos:
So if you have unit tests that need to test with different config values, which method do you use to get a fresh config?

I'm setting the new value via process.env.NODE_CONFIG and then deleting node config from the module cache and `require`'ing it again.

Issue #796, opened 2025-04-16
Split out util.js to a separate file


jdmarshall:
Config.get returns an object that still has access to all of the util features, which is why Config.util uses a lot of bind() calls in order to work. 

But some of the functions like loadFileConfigs() should probably be hoisted out of there, and possibly deprecated (recursive setModuleDefaults might make sense, but does loadFileConfigs()?)

At present if you call it, the returned values don't end up in the config data, however all of the local variables like NODE_CONFIG_PARSER, and configSources get updated which is wrong. 

There might be a better way to do all of this since 'this' isn't actually used that often. Perhaps something with a closure, as a replacement for the overbroad attachProtoDeep(). I'm not sure what the right solution is but what we have is not it. 

So it would be useful if someone could clever out how to split the shared state between config.js and the embedded util and extract it.

jdmarshall:
If a function in the new file took everything as arguments then this could satisfy #412 at the same time

jdmarshall:
Related: #728

jdmarshall:
toObject() seems to be the only function binding on this, and that could be split to have util return different data.

markstos:
I have not reviewed carefully recently but #571 is example of a re-think of the internals to serve as inspiration in whole or part.

jdmarshall:
It's been a while since I stared at util so I had forgotten a few bits. I think there are three sets of functions there.

1. Functions one expects to see on any object returned by Config.get() - those have to stay
2. Functions that generate the object that is returned by Config.get() - those could live anywhere
3. Functions for retrieving the data used to retrieve the data that is put into Config - and these should maybe live elsewhere

Most of 1 needs 'this'. Most of 2 don't or could have it passed in, like setImmutable does. 3 could be used for leveraging application specific adjustments to the load process. Especially for very large configuration trees.

PR #816, opened 2025-05-18
Extraction of the majority of the utilities from the config singleton

jdmarshall:
This PR contains a good deal of code to maintain feature and bug parity with the existing Config.util behavior while also extracting code that can be run either statically or as part of a separate workflow.

This is designed with elements of #569 in mind in addition to a number of other issues

- #815 
- #801 
- #806 
- #796 
- #753 

Providing workarounds for some and fixing a few outright.

This PR also includes code from #808 which this PR leverages.

These changes also include:

- almost 80 new unit tests 
  - including for functions that had no unit tests in Config.util such as `getOption()`
  - ordered to ensure that bugs in helper functions are reported before errors that use those functions
- some performance improvements
- a number of legibility/trace debugging ergonomic improvements for people investigating config issues
- a way to load a config directory with 0 side effects on Config
- careful support for calling this code recursively during a `require('config')` call - which substantially complicated the plumbing and necessitated the LoadInfo vs Env split.
- defer functionality pulled out, hopefully making it easier for someone to debug #687 
- support for CI/CD well-formedness checks enumerated across all environments 
 
To the best of my knowledge this contains no breaking changes, But it contains a number of TODOs that would be good candidates for a 4.1

Note that these commits also include `git mv` trickery to maintain commit history on both sides of the split files, to support forensics on future bug-vs-feature triage.

### DIFF
--- a/test/5-getConfigSources.js
+++ b/test/5-getConfigSources.js
@@ -15,7 +15,7 @@ vows.describe('Tests config.util.getConfigSources').addBatch({
       process.env.NODE_CONFIG = '{}';
       delete process.env.NODE_APP_INSTANCE;
       process.env.NODE_CONFIG_STRICT_MODE=0;
-      process.argv = ["node","path/to/some.js","--NODE_CONFIG='{}'"];
+      process.argv = ["node","path/to/some.js","--NODE_CONFIG='{}'"]; //TODO: This is a parse error so not testing the right thing
       var config = requireUncached(__dirname + '/../lib/config');
       return config.util.getConfigSources();
     },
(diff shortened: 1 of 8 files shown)


jdmarshall:
In retrospect we should have perhaps called 4.0.0 3.5.0 as this change is more of a 4.x change than the deprecations we removed.

jdmarshall:
config.js can shrink by around 400 lines once we start removing Config.util methods that are marked deprecated and have trampolines to lib/util.js. The duplicate jsdoc entries are about 40% of the moved code and are retained for documentation purposes.

test/5-getConfigSources.js
```diff
@@ -45,7 +45,7 @@ vows.describe('Tests config.util.getConfigSources').addBatch({
       return config.util.getConfigSources();
     },
 
-    'Two files should result in two entries': function(topic) {
+    'Three files should result in three entries': function(topic) {
         assert.equal(topic.length,3);
```
jdmarshall, line 49: This vexes me. Can we try to do better than this in the future?

markstos (approved):
Makes sense to me.

## QA Log

- Noted that the PR seems to pass existing tests, while *adding* more test coverage.

PR #818, opened 2025-05-29
Improve getPath performance

jdmarshall:
Fixes perf regression introduced by #808 and should improve single-step debugging experience.


jdmarshall:
This is, in retrospect, a fairly hot path so making it correct and fast is a useful goal.

PR #821, opened 2025-06-02
Remove files from configSources that are not part of the config sources.

### LINKED ISSUE #806: [BUG] util.loadFilesConfig() adds entries to configSource()

jdmarshall:
parseFile adding any calls to the configSources causes the union of the configSources to not match Config.get()

fixes #806

I don't want to merge this for a big yet, to give time for any other 4.0 issues to be sorted before we start merging more breaking changes. But I wanted to clear this off of my mental todo list.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -501,13 +501,9 @@ util.loadFileConfigs = function(configDir, options) {
     let opts = {...FIRST_LOAD.options, configDir, ...options};
     newLoad = new LoadInfo(opts);
     _load(newLoad);
-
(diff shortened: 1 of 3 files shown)


jdmarshall:
Coverage % drops slightly because this removed covered code from a file that has some uncovered functions in it. Probably should fix that in 4.0.x

PR #824, opened 2025-06-03
Simplify the merge logic in setModuleDefaults(), fix regression on trunk.

jdmarshall:
This code always did more work than it needed to but during the config.js/util.js split it sprouted a couple more defensive copies.

The existing code has some unnecessary tree merge logic in util.js, which makes the code difficult to follow. This change simplifies that code.

Within Config.util however, that 'unnecessary' logic does create one feature which my previous PR did not cover: Calling Config.get() prior to calling `setModuleDefaults` in 4.0 and earlier resulted in the Config object still being able to see the new defaults. That can only happen when allow_config_mutations is true, but this PR conserves that behavior when it is necessary.

### DIFF
--- a/test/2-config-test.js
+++ b/test/2-config-test.js
@@ -418,6 +418,25 @@ vows.describe('Test suite for node-config')
       assert.equal(moduleConfig.parm2, 2000);
     },
 
+    'Config.get() before setModuleDefaults() can see updates': function() {
+      process.env.ALLOW_CONFIG_MUTATIONS = true;
+
+      const mutableConfig = requireUncached(__dirname + '/../lib/config');
+
+      let defaults = {
+        someValue: "default"
+      };
+
+      try {
+        let customers = mutableConfig.get('Customers');
+
+        mutableConfig.util.setModuleDefaults('Customers', defaults);
+
+        assert.equal(customers.get("someValue"), "default");
+      } finally {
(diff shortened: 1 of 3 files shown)


test/2-config-test.js
```diff
@@ -418,6 +418,25 @@ vows.describe('Test suite for node-config')
       assert.equal(moduleConfig.parm2, 2000);
     },
 
+    'Preserves existing setModuleDefaults after Config.get() behavior': function() {
```
markstos: This is not a clear test name. Instead of referring to preserving existing behavior, say what the behavior is.  Explain what the purpose or behavior of calling setModuleDefaults after `config.get()` is.

If I understand, it's the behavior the config should be remain mutable via setModuleDefaults if `ALLOW_CONFIG_MUTATIONS` is true. 

Otherwise I trust you that this is a regression and agree that things should be mutable when that env var is set, so self-merge after the test name is updated.

markstos (changes requested):

test/2-config-test.js
jdmarshall: The problem is that I “simplified” a merge into a swap, which is only visible to the user if mutability is on. I never use mutability so I just saw overwrought object merges and removed all of them (turns out I can still remove most, except this one). 

I would like to propose mutable for the chopping block but only after we make deferred and setModuleDefaults work more cleanly.

PR #825, opened 2025-06-03
Make sure that multiple setModuleDefaults calls update the config sources properly

### LINKED ISSUE #827: [BUG] configSources contains incorrect data from multiple setModuleDefaults calls
This is the bug fixed by #825

jdmarshall:
I don't know if this constitutes a breaking change.  I don't know if anyone is relying on configSources for programmatic behavior.

Abundance of caution says yes.

fixes #827

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -964,7 +964,8 @@ class LoadInfo {
     }
 
     const path = moduleName.split('.');
-    Util.setPath(moduleDefaults, path, Util.cloneDeep(defaultProperties)); // TODO: This clobber is from the original code, and is a bug
+    const mergedDefaults = Util.extendDeep({}, defaultProperties, Util.getPath(moduleDefaults, path) || {}); // make sources match behavior - #822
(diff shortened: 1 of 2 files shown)


jdmarshall:
I'd like us to sort out the setModuleDefaults issue.

I think I may have a solution we could to in Util to keep the defaults and the loaded config as two separate pieces of data and do the heavy lifting in setModuleDefaults so that it's a zero cost abstraction if nobody is using this feature.

Since it's a startup only operation I'm a little less concerned about the perf. Plus I just landed some perf improvements anyway. 

I have another idea for fixing it but it might introduce a circular dependency so I need to think on it more:

```
const defaults = Util.loadFileConfigs({configDir: Path.join(__dirname, 'config')}); 
const myConfig = Config.withDefaults(defaults);
```

markstos (approved):
Looks OK, but let's not merge it ahead of the regression fix, since it's a breaking change and the regression fix may go out before 4.1.

## QA Log

- Reviewed diff
- Noted extra test coverage.

PR #826, opened 2025-06-12
Resolve setModuleDefaults in chronological order

jdmarshall:
If setModuleDefaults is called multiple times with overlapping values, make sure that they are accumulated in the correct order.

Fixes bug #825


markstos:
Was was this closed?

jdmarshall:
I got confused and had some branch fiddly problems. #828

PR #828, opened 2025-06-12
Apply overlapping setModuleDefaults calls in cronological order 

### LINKED ISSUE #822: [BUG] multiple calls to setModuleDefaults apply in the incorrect order
Because of the way we do the merge as a cumulative affair, a person might expect setModuleDefaults to treat these two as somewhat equivalent:

```
  Config.util.setModuleDefaults("a.b", settings1);
  Config.util.setModuleDefaults("a.b", settings2);

  // Expected behavior:

  let defaults = Config.util.extendDeep(settings1, settings2, Config.get("a.b"));
```

But this is what you actually get:

```
    let defaults = Config.util.extendDeep(settings2, settings1, Config.get("a.b"));
```

I don't see a particularly good way to solve this, and maybe that does recommend the solution in #670 of simply denying subsequent edits

See also #226

jdmarshall:
Fixes #822

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -459,6 +459,7 @@ class Util {
    * @param object {Object} - Object to set the property on
    * @param property {string|string[]} - The property name to get (as an array or '.' delimited string)
    * @param value {*} - value to set, ignoring null
+   * @return {*} - the given value
    */
   static setPath(object, property, value) {
     const path = Array.isArray(property) ? property : property.split('.');
@@ -480,6 +481,8 @@ class Util {
         next = next[name] = {};
       }
     }
+
+    return value;
   }
 
   /**
@@ -771,11 +774,13 @@ class Load {
    * @param options {LoadOptions=} - defaults to reading from environment variables
    */
   constructor(options, env = new Env()) {
-    this.config = {};
     this.env = env;
     this.options = { ...DEFAULT_OPTIONS, ...options };
     this.sources = this.options.skipConfigSources ? undefined : [];
     this.parser = this.options.parser;
+    this.config = {};
+    this.defaults = undefined;
+    this.unmerged = undefined;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


bin/node-config
```diff
@@ -0,0 +1,66 @@
...
+
+    console.log(`No errors for ${environment}`);
+  }
+}
+
+function dump(target) {
+  const env = process.env.CONFIG_NODE_ENV || process.env.NODE_ENV || 'default';
+
+  if (target === undefined) {
+    target = `${env}.json`;
```
markstos: Having an undocumented script that might overwrite production.json when you run it doesn't seem great. Why not dump to STDOUT and let the user handle where to write it?

bin/node-config
```diff
@@ -0,0 +1,66 @@
...
+  let loadInfo = LoadInfo.fromEnvironment(environment);
+
+  loadInfo.load();
+
+  return loadInfo.config;
+}
+
+function check(...environments) {
+  if (environments.length === 0) {
+    environments = ['production'];
```
markstos: Why take an environment as a CLI arg here but not NODE_APP_INSTANCE? And the API different for `dump`, which expects ENV vars, not CLI args.

I recommend using CLI args in all cases.

I'm not also not clear on the value of shipping this. There is already `node --check` to check of node code compiles, and dumping the config object seems easy enough.

I do find some value in a little CLI tool to look up a config value, though. I wrote one myself:

```bash
set -e

# Gather the techname
techname=$1; # First argument
configKey=$2; # Second argument

# default NODE_ENV to production
: ${NODE_ENV:=production} 

# Check that both are defined or throw an error.
: ${techname:?}
: ${configKey:?}

val=$(NODE_ENV=$NODE_ENV \
    NODE_APP_INSTANCE=$techname \
    node -p -e "\
      const val = require('config').get('$configKey');\
      if ((typeof val === 'string') || (typeof val === 'number')) {\
        val;\
      }\
      else {\
       JSON.stringify(val)\
     }"
)

# Check that val is defined or throw an error.
: ${val:?}

echo $val
```

lib/util.js
```diff
@@ -841,6 +845,28 @@ class LoadInfo {
     return this;
   }
 
+
+  /**
+   * scan and load config files in the same manner that config.js does
+   *
+   * @param loadInfo {LoadInfo}
```
markstos: a loadInfo param is defined here but doesn't appear to actually be used.

lib/util.js
```diff
@@ -841,6 +845,28 @@ class LoadInfo {
...
+    Util.loadFileConfigs(this);
+
+    if (additional) {
+      for (let {name, config} of additional) {
+        this.addConfig(name, config);
+      }
+    }
+
+    // Override with environment variables if there is a custom-environment-variables.EXT mapping file
+    this.loadCustomEnvVars();
```
markstos: 👍🏼 Seems like a good idea to move this to `util`.

markstos (changes requested):

bin/node-config
jdmarshall: It's documented in the wiki, but the link was not visible in the top level.

Also this functionality is part of the missing documentation for how to utilize some of the Util code directly, such as for loading module defaults via load()->setModuleDefaults(), which was not only covered in the initial extraction but also not possible due to the additional code I needed to move into Util, after breaking some of the circular top level variable usages in config.js.

>overwrite 

Generally config files are in the config directory not the project root.  But I could change it to write to stdout if no file is provided, sure.

The problem with dumping it to stdout is that you have to make sure the code absolutely never prints any other output to stdout. 

I've made some of your suggested changes on the cli PR.

bin/node-config
jdmarshall: > NODE_APP_INSTANCE

It does:

```
NODE_APP_INSTANCE=3 ../bin/node-config dump  | grep altDbPort
default -> stdout
    "altDbPort": 4400,
    "altDbPort1": 2209

```

I ran into situations where someone garbled the config files for one environment, and we started running checks for all of them in CI/CD. Deeper down in the code we are taking APP_INSTANCE (fromEnvironment()). We actually took it a step farther and validated every domain name we found in the app because we were getting gibberish due to people trying to do their own url interpolation which almost always ends badly.

The wages of sin on nearly-monolithic services without a monorepo are dark and desperate.

I also had a 'get' option in mind and I agree. I'm happy to add that later.

lib/util.js
jdmarshall: Ah, the original function did.

markstos:
This PR is still attempting to add a file to ./bin which is unrelated to the scope of "Apply overlapping setModuleDefaults calls in cronological order". Perhaps it needs to be rebased.

jdmarshall:
I cannot figure out what's going on with this anymore and github is showing actionable items that it won't link to.

PR #830, opened 2025-06-12
Cli

jdmarshall:


jdmarshall:
This also fixes a bug with skipConfigSources && setModuleDefaults()

lib/util.js
```diff
@@ -841,6 +840,27 @@ class LoadInfo {
     return this;
   }
 
+
+  /**
+   * scan and load config files in the same manner that config.js does
+   *
+   * @param additional {{name, value}[]=} additional values to populate (usually from NODE_CONFIG
```
markstos, line 873: There's a missing trailing ")" here.

lib/config.js
```diff
@@ -575,7 +554,7 @@ function _init(loadInfo) {
   // Place the mixed NODE_CONFIG into the environment
   loadInfo.setEnv('NODE_CONFIG', JSON.stringify(Util.extendDeep(envConfig, cmdLineConfig, {})));
 
-  _load(loadInfo, additional);
+  loadInfo.load(additional);
```
markstos: If we haven't released it yet, let's reconsider including "Info" in the name of anything. Like "data", the word "info" doesn't really add value, since all variables are "info" and "data". It's not clear for example how "info" is different from "config" how returning "LoadInfo" is different from returning a config. "load" or some other more precise name would be better.

jdmarshall: Then you're going to have Load as a noun and a verb: `Load.load()`. I don't think that's better. I'm open to other suggestions though. Model, Load, and Data are the only ones that come to mind even with the help of a thesaurus. And if you don't like Info and data I don't think you're going to like model, job, or run much better? (I confess I'd prefer a word that is not both a noun and a verb but an adjective is as far as I got)

We already overload `config` btw. There are places in config.js where 'config' is an instanceof Config and places where it starts as an Object, and never becomes a Config. Part of these changes is deleting some of those.

jdmarshall:
Did I lose you?

jdmarshall:
This relates to work being done to satisfy #569

jdmarshall:
> Haven't needed that in the last 10 years.

How many devs do you have touching configuration? We had three different teams and even the platform team occasionally made mistakes.

Resilience comes in how you recover from mistakes, robustness is just about how you prevent them.

PR #834, opened 2025-06-18
BugFix: shared references could trigger "Cannot redefine property"

markstos:
Under certain circumstances in a config with a shared reference,
makeImmutable could end up being called twice on the same property,
throwing an error.

As a fix, we check if a property is already immutable, and if so, move on.

## QA Log

- Added test coverage, confirming that the the test failed before the patch and passes after it.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -385,11 +385,15 @@ util.makeImmutable = function(object, property, value) {
         })
       }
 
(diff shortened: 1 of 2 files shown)


jdmarshall:
This function is moved to lib/util.js in a couple of different open PRs. 

It might also do better as Util.isWritable()

Issue #687, opened 2022-08-06
Deferred values don't work with sub-modules


EricLeeSanders:
### I'm submitting a ... 
  - [ x ] bug report
  - [  ] feature request
  - [  ] support request or question => Please do not submit support request or questions here, see note at the top of this template.

### What is the current behavior?

Sub-modules that use node-config are not able to reference `deferred` values from the parent module.

For example, say you have a `parent` module with `node-config` installed and two config files, `default` and `production`.

**default.js**
```
const defer = require('config/defer').deferConfig;

module.exports = {
  firstName : 'default first name',
  lastName: 'default last name',
  fullName : defer(function ()  {
    return this.firstName+' '+this.lastName;
  })
} 
```

**production.js**
```
module.exports = {
    firstName : "Jane",
    lastName: "Doe",
 }
```

**index.js**
```
const config = require('config');
console.log('from parent: ' + config.fullName);
```

This correctly prints the full name, `Jane Doe`.

However, say I also have a `sub-module` with the following index file

**child - index.js**
```
const config = require('config');
console.log(config);
console.log('from sub-module: ' + config.fullName);
```

Then I update the `parent module` index file to the following to require the child module.
**parent - index.js**
```
const config = require('config');
console.log('from parent: ' + config.fullName);

const configDeferTestChild = require('config-defer-test-child');
```

The output is then:
```
from parent: Jane Doe
Config {
  firstName: 'Jane',
  lastName: 'Doe',
  fullName: DeferredConfig {
    prepare: [Function (anonymous)],
    resolve: [Function (anonymous)]
  }
}
from sub-module: [object Object]
```

Which you can see that the config values correctly resolve the values in the `parent` module. But in the `sub-module` it does not resolve. 

### What is the expected behavior?

Based on the wiki, I would expect that `config-defer` should work with sub modules as well and I should see "Jane Doe" in the sub module index file (I haven't read anything that stated otherwise).

### Please tell us about your environment:

 -  node-config version: 3.3.7
 -  node-version: 18.0.0

markstos:
I have not looked closely to see if this possible to implement. 

You can submit a PR to find out what's involved and we can discuss any trade-offs from there. I don't think anyone else is going to implement this for you, though.

markstos:
In your example, did you also test with `config.get('fullName')`?

I'm not sure this issue can be resolved without perhaps a re-design that introduces some incompatibilities. See for example the proposal in #569.

speedytwenty:
[...]

PR #838, opened 2025-07-15
Segregate defaults data from config data, so deferreds and multiple setModuleDefaults calls can be applied properly.

jdmarshall:
This fixes:
 - #687 
 - #822
 - #827

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -459,6 +459,7 @@ class Util {
    * @param object {Object} - Object to set the property on
    * @param property {string|string[]} - The property name to get (as an array or '.' delimited string)
    * @param value {*} - value to set, ignoring null
+   * @return {*} - the given value
    */
   static setPath(object, property, value) {
     const path = Array.isArray(property) ? property : property.split('.');
@@ -480,6 +481,8 @@ class Util {
         next = next[name] = {};
       }
     }
+
+    return value;
   }
 
   /**
--- a/lib/util.js
+++ b/lib/util.js
@@ -774,11 +774,13 @@ class Load {
    * @param options {LoadOptions=} - defaults to reading from environment variables
    */
   constructor(options, env = new Env()) {
-    this.config = {};
     this.env = env;
     this.options = { ...DEFAULT_OPTIONS, ...options };
     this.sources = this.options.skipConfigSources ? undefined : [];
     this.parser = this.options.parser;
+    this.config = {};
+    this.defaults = undefined;
+    this.unmerged = undefined;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jdmarshall:
re-filing this one.

Now if I can just locate the resolveDeferred fix I had that built on top of this...

jdmarshall:
This is a small breaking change but still breaking, if you use config.sources() to try to make decisions.

markstos (approved):
It's hard to see from exactly what's happening from the diff, but from test coverage I can see this isn't really breaking anything and it's clearly fixing some things.

Thanks for knocking down some longstanding bugs @jdmarshall !

## QA Log

- Reviewed diff
- Reviewed test coverage

lib/util.js
```diff
@@ -1022,24 +1027,27 @@ class Load {
    * @return {Object} - The module level configuration object.
    */
   setModuleDefaults(moduleName, defaultProperties) {
-    const moduleConfig = Util.cloneDeep(defaultProperties);
-    let moduleDefaults;
+    if (this.defaults === undefined) {
```
jdmarshall, line 1030: I’ll add some explanations for posterity since this code seems to be a bit of a chore to read. 

This is a zero cost abstraction. To fix the bugs required more memory footprint for config. So don’t allocate the extra storage if the feature is not being used.

lib/util.js
```diff
@@ -1022,24 +1027,27 @@ class Load {
...
 
-    if (this.sources === undefined) {
-      moduleDefaults = {};
-    } else if (this.sources.length > 0 && this.sources[0].name === 'Module Defaults') {
-      moduleDefaults = this.sources[0].parsed;
-    } else {
-      let source = { name: 'Module Defaults', parsed: {} };
-      this.sources.splice(0, 0, source);
-      moduleDefaults = source.parsed;
+      if (this.sources) {
```
jdmarshall, line 1034: We only need to set the source on the first call, which we now have a convenient way to do instead of trying to splice it in on every set call.

lib/util.js
```diff
@@ -1022,24 +1027,27 @@ class Load {
...
+        this.sources.splice(0, 0, { name: 'Module Defaults', parsed: this.defaults });
+      }
     }
 
     const path = moduleName.split('.');
-    Util.setPath(moduleDefaults, path, Util.cloneDeep(defaultProperties)); // TODO: This clobber is from the original code, and is a bug
+    const defaults = Util.setPath(this.defaults, path, Util.getPath(this.defaults, path) ?? {});
```
jdmarshall, line 1040: Set if missing, as I’m the original code

lib/util.js
```diff
@@ -1022,24 +1027,27 @@ class Load {
...
+
+    Util.extendDeep(defaults, defaultProperties);
 
-    Util.extendDeep(moduleConfig, Util.getPath(this.config, path) || {});
+    const original =
```
jdmarshall, line 1044: Snapshot the config data as it existed before we added defaults to it, so that duplicate calls to set defaults can be resolved in chronological order.

lib/util.js
```diff
@@ -1022,24 +1027,27 @@ class Load {
...
+      Util.getPath(this.unmerged, path) ??
+      Util.setPath(this.unmerged, path, Util.getPath(this.config, path) ?? {});
+
+    const moduleConfig = Util.extendDeep({}, defaults, original);
```
jdmarshall, line 1048: Only clone the config for this single path, since it’s expensive.

test/0-util.js
```diff
@@ -483,25 +487,60 @@ vows.describe('Tests for util functions')
       }
     },
     'Load.setModuleDefaults()': {
```
jdmarshall, line 489: These testing library only runs the setup function once and shares values between tests. These tests were not testing what they claimed to test, which is probably how the bugs escaped into trunk in the first place.

test/0-util.js
```diff
@@ -483,25 +487,60 @@ vows.describe('Tests for util functions')
...
         assert.deepEqual(load.config, { foo: { field1: 'set', field2: 'another' } });
       },
-      'can be called multiple times for the same key': function (load) {
+      'getSources() and Config.get() are consistent with each other': function () {
+        let loadInfo = new Load({});
+
+        loadInfo.setModuleDefaults("foo", { field2: 'another', field4: "1"});
+        loadInfo.setModuleDefaults("foo", { field3: 'additional', field4: "2"});
+
+        // this is probably a bug (#822) but at least the sources and the config now agree with each other.
```
jdmarshall, line 508: This comment is now incorrect.

jdmarshall:
Since we don’t do CI for releases, I can still push one out. But perhaps we should close that hole.

PR #845, opened 2025-07-17
Delete deprecated functions in Config.util, and associated tests.

jdmarshall:
This finishes the deprecation warnings introduced in 4.0

This should sit a while.


jdmarshall:
@markstos anything else you’d like to put into a 4.1 release?

markstos:
@jdmarshall Nothing else, thanks.

markstos (approved):
Approved as a natural progression after 4.0 deprecation warning.

jdmarshall:
Rebased on master.

erunion:
I know I am shouting into the void here but I would like to voice my displeasure at the removal of this in a minor release. Yes I am aware that the API was deprecated in v4 but APIs should be removed in major releases. It's a major release!

Anybody that is using this and has `^4.0.0` installed will see breakages if any dependency they're using has `^4.0.0` as a dependency of its own.

PR #847, opened 2025-07-18
Switch to c8 for code coverage.

jdmarshall:
c8 is compatible with nyc but uses nodejs' internal tooling for code coverage. 

One of the consequences of switching to this is that we no longer need to filter out the .ts tests from coverage, as it seems not to interact with require.extensions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,11 +26,11 @@
   },
   "devDependencies": {
     "@types/node": "^7.0.8",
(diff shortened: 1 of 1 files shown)


markstos (approved):
Let's try it.

jdmarshall:
Found it after a breakage ok another project. It uses v8”s profiling code.

Issue #841, opened 2025-07-15
ESM support: Roadmap to 5.0


jdmarshall:
## Summary

Some initial brainstorming ideas around migrating `node-config` to ESM and potentially eliminating the singleton entirely.
Given that all of the LTS supported NodeJS versions now have a way to require ESM modules, and we now [support strict](#815) one or both of these seemed reasonable, even overdue.

Subsequently, I ran into a number of roadblocks, particularly to do with how idiomatic use of `require` inside of the config directly is very much at odds with how ESM modules want to handle imports. This leads to issues with `createRequire` even before we get into the tests, which themselves are also actively antagonistic to use of `import`, `createRequire`, or any mixture of both.

In the interests of regaining some momentum, I am now cutting scope a bit and altering the strategy. I have a route forward that I believe gives us more choices going forward for the features that are being cut, while giving me a straighter path to a 5.0 release that I can ship.

This work is still informed by [#569](https://github.com/node-config/node-config/issues/569), with some continued additions, alterations, and substitutions, informed by other feature requests and bug reports.

## The Plan

For 4.3 I am working toward the following goals:

[x] reduce reliance on requireUncached in the tests on this function - previously 80 uses. 
[x] introduce `defer` and `util` to executable config files by supporting export of a callback function to evaluate the file
[x] deprecate async.js 
[x] relocate deferConfig's implementation into lib, and deprecate node-config/defer, which doesn't work great with imports anyway
[x] modify `defer` to accept async functions
[ ] deprecate raw.js
[ ] expose raw through the same callback mechanism as defer

For 5.0 then,

 - Utilize a cache busting workaround for the remaining uses of requireUncached
 - Remove async.js, defer.js, raw.js
 - Remove `config.util.makeImmutable()`
 - Fix immutability bugs in lib/util.js
 - Switch to ESM
 - use `createRequire` to replace `require` in `loadFiles()` and elsewhere
 - Create prerelease builds of 5.0.0 for public commentary.

### Bootstrapping issues

The `require()` thunking layers don't like exporting a module that is still being loaded. 

The most straightforward way to fix this also addresses #740, which is to modify node-config to allow executable config
files to return a function instead of the data. The function can then receive `defer` and `util` as arguments instead of having to load them. This is a small but 

### Imports are Immutable

If you look at the requireCache for createRequire you would think that there's a way to reset ESM modules the same way you reset CJS modules for testing purposes. You would be wrong. So we need a way to completely reset the state of node-config for different test suites.

[...]

PR #850, opened 2025-07-29
feat: add rollup build config for support esm #841

liaohui5:
1.add rollup.config.js and rollup plugins
2.updated package.json, add exports fields and build command
3.updated `.gitignore`

## Summary by CodeRabbit

* **Chores**
  * Updated project configuration to improve build process and compatibility for both ES module and CommonJS consumers.
  * Added new build script and development dependencies for bundling.
  * Enhanced `.gitignore` to exclude build output and system files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ npm-debug.log
 
 .nyc_output
 coverage
+
+dist
+.DS_Store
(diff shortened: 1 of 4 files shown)


jdmarshall:
I would rather pursue direct ESM support than add a cross compilation tool to this code. My recent experience dealing with these things has not been pleasant. 

There are now ways to load ESM modules into LTS node versions. As I’ve said in #841 i think that’s the way forward and it’s now better to jump than to waddle on via cross compilers.

Issue #860, opened 2026-01-09
Reduce usage of requireUncached in tests


jdmarshall:
Make sure any tests that can use util.js are using it, to reduce the number of tests relying on requireUncached.

PR #862, opened 2026-01-10
Replace some integration tests with unit tests to reduce reliance on `requireUncached`

jdmarshall:
Work for #860

This changeset fixes a silently broken test for Unicode BOM characters, adds coverage for the optional parameter to `fromEnvironment()`, and moves a heap of tests still being run on the Config singleton to run on loadFile() or loadFileConfigs() as appropriate.  A few of these were already doing this but had been modified in-place. 

This PR reduces the number of uses of `requireUncached` in the tests from 80 to 46. I believe all of the remaining cases are
for functionality that was not migrated to lib/util.js

## Summary by CodeRabbit

* **Configuration Changes**
  * Modified NODE_CONFIG_DIR handling to no longer automatically convert directory paths to absolute paths; directory values are now used as provided

* **Tests**
  * Consolidated and expanded test coverage with comprehensive scenarios for configuration loading, environment variable handling, multiple file formats, and directory resolution

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


lib/util.js
```diff
@@ -1158,8 +1158,6 @@ class Load {
     }
 
     let configDir = env.initParam('NODE_CONFIG_DIR');
-    configDir = configDir && _toAbsolutePath(configDir);
```
jdmarshall, line 1161: This code is 1) incorrect because this can be a colon-separated value, and 2) duplicates work that is repeated in the next phase of the load operation anyway, after the paths have been split for iterating.

PR #864, opened 2026-01-17
withModuleDefaults()

jdmarshall:
This is a PR to address #572

This should not be landed until 4.3 is released.

Hide Whitespace is recommended, since I turned Config and ConfigUtil into classes, and also made `setModuleDefaults` and `withModuleDefaults` sibling test suites and indented the old tests by one. Most of the new tests are copies of the `setModuleDefaults` tests but asserting that the changes are not shared back to the original object.

This is around 160 lines of new tests, 30 lines of JSDoc, and moving an unexported 30 line function out of the middle of a block of exposed methods. The rest is new code.

This PR ended up being a good time to try again to convert Config to a class, which was a bit stranger than I expected it to be. Some fun infinite loop issues when I enumerated the changed set of properties caused by not mangling the object prototype as broadly as the original implementation does. 

I also had to get rid of two more file-scope variables to make `withModuleDefaults` work, since the whole point of the new instance is to create a different loadout than the original.

## Summary by CodeRabbit

* **New Features**
  * Config now supports per-instance contexts with validated get/has, optional immutability, per-module defaults (in-place and non-destructive), reloadable file-based configs, deep prototype wiring, exportable serialized configs, visible config sources, runtime strictness checks, and environment/load snapshots.

* **Tests**
  * Expanded coverage for cloning/snapshots, immutability enforcement, module defaults, non-destructive defaults, and related regressions.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -20,6 +20,12 @@ const DEFAULT_CLONE_DEPTH = 20;
  * @class
  */
 class ConfigClass {
+  /**
+   * Non-enumerable field for util functions
+   *
+   * @type {ConfigUtils}
+   */
+  util;
 
   /**
    * <p>Get the configuration object.</p>
@@ -57,7 +63,7 @@ class ConfigClass {
     this[LOAD_SYMBOL] = load;
     Util.extendDeep(this, load.config);
 
-    this.util = new ConfigUtils(this);
+    Util.makeHidden(this, 'util', new ConfigUtils(this));
     this.util.attachProtoDeep(this);
     // Perform strictness checks and possibly throw an exception.
     this.util.runStrictnessChecks(this);
--- a/test/2-config-test.js
+++ b/test/2-config-test.js
@@ -56,6 +56,10 @@ describe('Test suite for node-config', function() {
     it('Config extensions are included with the library', function() {
       assert.strictEqual(typeof config.util.makeImmutable, 'function');
     });
+
+    it('config util is not an enumerable property', function() {
+      assert.strictEqual(Object.keys(config).includes('util'), false, 'util in Object.keys()');
+    });
   });
 
   describe('Immutability', function() {


lib/config.js
```diff
@@ -55,402 +56,564 @@ let checkMutability = true;      // Check for mutability/immutability on first g
...
+   * The above example results in a "MyModule" element of the configuration
+   * object, containing an object with the specified default values.
+   * </p>
+   *
+   * @method setModuleDefaults
+   * @param moduleName {string} - Name of your module.
+   * @param defaultProperties {Object} - The default module configuration.
+   * @return moduleConfig {Object} - The module level configuration object.
+   */
+  setModuleDefaults(moduleName, defaultProperties) {
```
jdmarshall, line 193: I don't like this. Thought I fixed this but I must have missed a function.

lib/config.js
```diff
@@ -55,402 +56,567 @@ let checkMutability = true;      // Check for mutability/immutability on first g
...
+  /**
+   * @deprecated please use Parser.stripYamlComments
+   * @param fileStr
+   */
+  stripYamlComments(fileStr) {
+    return this.#config[LOAD_SYMBOL].parser.stripYamlComments(fileStr);
+  }
 }
 
 /**
```
jdmarshall, line 626: I had to move this down. No changes made.

lib/config.js
```diff
@@ -497,131 +663,21 @@ function _init(load) {
...
-  }
+module.exports = (() => {
+  const load = Load.fromEnvironment();
 
-  // Recursion detection
-  depth = (depth === null ? DEFAULT_CLONE_DEPTH : depth);
-  if (depth < 0) {
-    return toObject;
-  }
+  _init(load);
```
jdmarshall, line 674: This is now the last block of code that still forces this to be a singleton.

test/0-util.js
```diff
@@ -1256,5 +1256,39 @@ describe('Tests for util functions', function () {
...
+      assert.strictEqual(typeof load.clone, 'function');
+    });
+
+    it('copies data', function () {
+      let load = new Load({configDir: __dirname + '/config'});
+      load.scan();
+
+      let newLoad = load.clone();
+
+      assert.strictEqual(typeof newLoad.config.Customers, 'object');
```
jdmarshall, line 1674: I wanted to do a deep comparison here but it turns out the `random` example in test/config thwarts this, since it gets evaluated every time and the value changes.

jdmarshall:
Huh. Score one for CodeRabbit.

lib/config.js
```diff
@@ -55,402 +56,567 @@ let checkMutability = true;      // Check for mutability/immutability on first g
...
+      throw new Error("Calling config.get with null or undefined argument");
+    }
 
-  // Perform strictness checks and possibly throw an exception.
-  util.runStrictnessChecks(t);
-};
+    // Make configurations immutable after first get (unless disabled)
+    if (checkMutability) {
+      if (!this[LOAD_SYMBOL].initParam('ALLOW_CONFIG_MUTATIONS', false)) {
+        this.util.makeImmutable(this);
```
jdmarshall: This is a bug. #866 adds a test which will mark this as a regression.

lib/config.js
```diff
@@ -55,402 +55,566 @@ let checkMutability = true;      // Check for mutability/immutability on first g
...
+
+    load.options.nodeEnv.forEach(function (env) {
+      // Throw an exception if there's no explicit config file for NODE_ENV
+      const anyFilesMatchEnv = sourceFilenames.some(function (filename) {
+        return filename.match(env);
+      });
+      // development is special-cased because it's the default value
+      if (env && (env !== 'development') && !anyFilesMatchEnv) {
+        _warnOrThrow(`${load.getEnv("nodeEnv")} value of '${env}' did not match any deployment config file names.`);
+      }
```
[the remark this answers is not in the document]
jdmarshall: Hmm.  This is an existing bug, but I think it's in service of allowing test-1.js to suffice. 

I'll file a separate ticket to expose this. #867

jdmarshall:
I would like to land this one also because working on this feature finally revealed to me that makeImmutable can in fact be moved to util. The parts tying it to config.util are all duplicate effort that I can now remove and finally get the rest of the functionality moved over to util.js.

markstos (approved):
I see the use of this. Thanks for the work.

test/2-config-test.js
```diff
@@ -307,100 +330,225 @@ describe('Test suite for node-config', function() {
...
-      var testSubModuleConfig = config.get('services');
+        let testModuleConfig = newConfig.get('BKTestModule');
+        var testSubModuleConfig = newConfig.get('services');
 
-      assert.deepEqual(BKTestModuleDefaults.nested, testModuleConfig.get('nested'));
-      assert.deepEqual(OtherTestModuleDefaults.other, testSubModuleConfig.OtherTestModule.other);
+        assert.deepEqual(BKTestModuleDefaults.nested, testModuleConfig.get('nested'));
+        assert.deepEqual(OtherTestModuleDefaults.other, testSubModuleConfig.OtherTestModule.other);
+      });
     });
```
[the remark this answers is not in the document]
jdmarshall: Please review this code again.

Issue #865, opened 2026-01-19
Immutability checks fail to guarantee immutability


jdmarshall:
makeImmutable only makes the children immutable, not the given object

```
    it('Correctly unable to change an immutable configuration', function() {
      config = requireUncached(__dirname + '/../lib/config');

      config.util.makeImmutable(config, 'TestModule');

      assert.throws(() => config.TestModule.newField = "setToThis",
        /Cannot assign to read only property/);
    });
```

```
  AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Missing expected exception.
      at TestContext.<anonymous> (/Users/jasonmarshall/Projects/cobbler/node-config/test/2-config-test.js:282:14)
      at Test.runInAsyncScope (node:async_hooks:214:14)
      at Test.run (node:internal/test_runner/test:1047:25)
      at Suite.processPendingSubtests (node:internal/test_runner/test:744:7)
   {
    generatedMessage: false,
    code: 'ERR_ASSERTION',
    actual: undefined,
    expected: /Cannot assign to read only property/,
    operator: 'throws'
  }
```

This was partially fixed for #505 but only for the children and not the object itself. I think the recursion is done wrong here.

But it turns out there are bigger problems. 

If your first call to Config looks like:

```
config.Customers.get('dbString')
```

Then only config.Customers will be marked as immutable (due to attachDeep and 'this' behavior).

Any per-object tracking I might want to do is then hampered by the fact that you cannot set a property on an object that's already marked as immutable to determine if it needs to be marked as immutable.  So that needs some work.

jdmarshall:
This is actually the least of the problems with makeImmutable().

It turns out the check will misfire and not mark most of the Config as immutable if the first use looks like:

```
config.Customers.get('dbString')
```

Customers will be marked (mostly) as Immutable, the flag will be cleared, and subsequent calls to other fields will then skip the immutability check because it's already set as false. 

These are the wages of making simple, single-field tests for complex, recursive features.

jdmarshall:
Parts of this are fixed in the withModuleDefaults functionality but not all.

PR #866, opened 2026-01-20
Prevent extension when calling makeImmutable(), and add pinning tests

jdmarshall:
Ask JavaScript if the object is mutable instead of trying to track state

Fix for both bugs in #865, and makes #864 more agreeable to honoring the immutability contract.

**Note** while this PR is marked as a breaking change, absolutely none of our existing tests failed with this change,
meaning we have documented behavior that we do not test for.

Looking into the commit history, the contrary comment in the JSDoc dates to a time when this function was about
eight lines long and had none of the other many, many additions to enforce immutability. I propose that this is not actually a useful feature anymore. But I will mark this as 4.4 which will have several other breaking changes, to leave 4.3 more or less in peace. 

## Summary by CodeRabbit

* **Refactor**
  * Updated immutability behavior to use a runtime extensibility check and added partial-mode so selected top-level properties can remain extensible while nested settings stay immutable.

* **Tests**
  * Added tests ensuring immutability holds after sequential accesses and that adding new properties to an immutable configuration is prevented.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -9,7 +9,6 @@ const { Util, Load } = require('./util.js');
 const Path = require('path');
 
 const DEFAULT_CLONE_DEPTH = 20;
-let checkMutability = true;      // Check for mutability/immutability on first get
 
 /**
  * <p>Application Configurations</p>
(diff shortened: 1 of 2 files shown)


jdmarshall:
> setModuleDefaults with new modules after get() will fail with immutability enabled.

That's documented as unsupported. Where do you see the opposite mentioned in the docs?

> the JSDoc further down (around lines 233-235 per the past review) still states that new properties can be added, which contradicts the new Object.preventExtensions behavior in full mode.

Now you're hallucinating. 233-235 has always been an empty paragraph. Unless there's some whitespace morse code I'm not seeing...

Issue #886, opened 2026-02-19
Performance regression in .get() after 4.3


jdmarshall:
```
access functions ⇒ get()

 ⇒ 3.3.12                  ▏███████████████████▌▕ 1,897,304 op/s | 36 samples | (baseline)
 ⇒ 4.0.1                   ▏███████████████████▌▕ 1,896,264 op/s | 39 samples | (1.00x slower)
 ⇒ 4.2.0                   ▏███████████████████▌▕ 1,900,317 op/s | 38 samples | (1.00x faster)
 ⇒ trunk                   ▏████████████████────▕ 1,580,743 op/s | 39 samples | (0.83x slower)
 ⇒ branch                  ▏████████████████████▕ 1,917,092 op/s | 38 samples | (1.01x faster)
```

jdmarshall:
63f924d7d58f0ddfb45af4d78b500e3e79dd21c5 is the problem.

PR #887, opened 2026-02-19
Matrix builds and separate benchmarks from ci run.

jdmarshall:
I'm missing the benchmark results in the PR listing, and also with recent changes we should absolutely be making sure the code still runs properly on node 20 and 22.

So, this change runs the type checking and unit tests on 20, 22, and 24, but as a separate step runs the benchmarks as a separate task (on node 24) for better visibility. 

This is closer to the way some other projects I work with function.

## Summary by CodeRabbit

* **Chores**
  * CI now runs unit tests on Node.js 22.x and 24.x with dependency caching.
  * CI Node setup action updated to a newer version.
  * Added a separate benchmarks job to the workflow.
  * Pipeline now performs a dedicated type-check step before running tests.
  * Test invocation simplified for more reliable test runs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,17 +14,41 @@ jobs:
   Unit-Tests:
     runs-on: ubuntu-latest
     timeout-minutes: 10
+
+    strategy:
+      matrix:
+        node-version: [ 22.x, 24.x ]
+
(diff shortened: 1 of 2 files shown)


jdmarshall:
Motivation: #886

PR #902, opened 2026-03-09
Icing .iced files

jdmarshall:
.iced files have never worked, not once, since introduction. And there are no tests to assert that it does. A function that never worked cannot be a breaking change, so this isn't.

Addresses #900

This should ever so slightly improve the speed of scan() and thus library initialization.

## Summary by CodeRabbit

* **Chores**
  * Removed iced-coffee-script parser support. The parser will no longer recognize or process files of this type.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -20,7 +20,6 @@ let Yaml = null,
 // Define soft dependencies so transpilers don't include everything
 let COFFEE_2_DEP = 'coffeescript',
     COFFEE_DEP = 'coffee-script',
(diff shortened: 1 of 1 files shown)


markstos (approved):
## QA Log

- Reviewed diff
- Noted there is nothing in package.json to remove.

jdmarshall:
I kept going through the git history because surely this was a migration error but no. 

The giveaway was no tests. Which just reinforces the policy of asking for tests in additive PRs.

PR #915, opened 2026-06-09
Don't install dependencies younger than a week old.

jdmarshall:
This flag is meant to prevent supply chain attacks in compatible npm versions.

## Summary by CodeRabbit

* **Chores**
  * Release configuration updated to set the minimum release age to 7 (adjusts timing controls for releases). Existing registry and publish safeguards remain unchanged. This increases the minimum time before a release is eligible, helping reduce accidental or premature releases.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 # Prevent accidental publish to private repository
 registry=https://registry.npmjs.org/
+minimumReleaseAge=7


markstos:
I recommend 7 days, providing a larger buffer for vulns to be discovered in recent releases. It can always be overridden if there's some hot security fix to incorporate.

PR #916, opened 2026-06-10
4.4 branch: Guard against odd object prototypes triggered by changes in TOML 4.0

### LINKED ISSUE #911: [BUG] target.hasOwnProperty is not a function when using toml
**Describe the bug**

When reading a property of an object from the config with toml a target.hasOwnProperty is not a function error occurs.

config/default.toml:
```toml
[foo]
bar = "value"
```

test.js:
```js
const config = require("config");

console.log(config.has("foo.bar"));
console.log(config.get("foo.bar"));
```

output:
```
true
/home/test/node_modules/config/lib/util.js:191
              if (target.hasOwnProperty(property) || (property in target && typeof target[property] !== 'function')) {
                         ^

TypeError: target.hasOwnProperty is not a function
    at Object.get (/home/test/node_modules/config/lib/util.js:191:26)
    at Util.getPath (/home/test/node_modules/config/lib/util.js:606:25)
    at ConfigClass.get (/home/test/node_modules/config/lib/config.js:101:24)
    at Object.<anonymous> (/home/test/test.js:4:20)
    at Module._compile (node:internal/modules/cjs/loader:1761:14)
    at Object..js (node:internal/modules/cjs/loader:1893:10)
    at Module.load (node:internal/modules/cjs/loader:1481:32)
    at Module._load (node:internal/modules/cjs/loader:1300:12)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:245:24)

Node.js v24.13.0
```

**Expected behavior**
The property should be read.

 -  node-config version: 4.4.1
 -  node-version: 24.13.0
 - toml version: 4.1.1
 
### Other information

Works with config v4.3.0 and stops working with v4.4.0

jdmarshall:
Note that test in 0-util.js:

    assert.deepStrictEqual(result.config.messages, [...

Still fails with toml 4.1 because there is no prototype on messages, which is problematic

Fixes #911 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved object immutability property detection to use more reliable methods.

* **Chores**
  * Updated continuous integration workflow to include additional branch for automated testing.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -188,7 +188,7 @@ class Util {
             get(target, property, receiver) {
               // Bypass proxy receiver for properties directly on the target (e.g., RegExp.prototype.source)
               // or properties that are not functions to prevent errors related to internal object methods.
-              if (target.hasOwnProperty(property) || (property in target && typeof target[property] !== 'function')) {
+              if (Object.hasOwn(target, property) || (property in target && typeof target[property] !== 'function')) {
                 return Reflect.get(target, property);
               }
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,6 +5,7 @@ on:
     types: [opened, synchronize, reopened]
     branches:
       - master
+      - '4.4'
 
 concurrency:
   group: pr-checks-${{ github.head_ref }}


lib/util.js
```diff
@@ -186,9 +186,13 @@ class Util {
...
           value = new Proxy(this.makeImmutable(value), {
             get(target, property, receiver) {
+              const ownProps = [
+                ...Object.getOwnPropertyNames(target),
+              ]
+
               // Bypass proxy receiver for properties directly on the target (e.g., RegExp.prototype.source)
               // or properties that are not functions to prevent errors related to internal object methods.
-              if (target.hasOwnProperty(property) || (property in target && typeof target[property] !== 'function')) {
+              if (Object.hasOwn(target, property) || (property in target && typeof target[property] !== 'function')) {
```
jdmarshall, line 191: So I've been wondering if the 'property in' is just really old code but it turns out that branch triggers for 'constructor' because target.hasOwnProperty('constructor') fails for many objects. This could still probably use a bit of love.

PR #926, opened 2026-08-12
Fix resolveAsyncConfigs leaving promises in the exported config

### LINKED ISSUE #925: v5: config.get() still returns a Promise after Util.resolveAsyncConfigs — documented async pattern has no working equivalent
* Used fable to provide this issue / pr -- have not vetted the PR, so take it or leave it :)

### Summary

The documented migration path for async configuration values ([wiki: Promises and async functions](https://github.com/node-config/node-config/wiki/Special-features-for-JavaScript-configuration-files#promises-and-async-functions)) does not work in v5.0.0: after `Util.resolveAsyncConfigs(config)` settles, `config.get()` still returns the `Promise`, not the resolved value.

Since v5 also removed `config/async.js` (whose `asyncConfig` **did** correctly replace promises with values in v4.x), there appears to be no working equivalent for "resolve async values, then read them synchronously via `config.get()`" in v5.

### Repro

```
mkdir -p repro/config && cd repro
npm init -y && npm install config@5.0.0
```

`config/default.js` (taken from the wiki pattern):

```js
async function dbSecrets() {
  await new Promise((r) => setTimeout(r, 20));
  return 'secret-from-async';
}

module.exports = ({ defer }) => ({
  mode: 'dev',
  fromDatabase: defer(dbSecrets),
});
```

`index.js` (taken from the wiki's "Resolving async configurations" section):

```js
const { Util } = require('config/lib/util');
const config = require('config');

Util.resolveAsyncConfigs(config).then(() => {
  const value = config.get('fromDatabase');
  console.log('config.get():', value);
  console.log('is Promise:', value instanceof Promise);
});
```

```
$ node index.js
config.get(): Promise { 'secret-from-async' }
is Promise: true
```

**Expected** (per the wiki): `config.get('fromDatabase')` returns `'secret-from-async'`.

**Actual:** it returns the (settled) `Promise`. Every consumer would have to `await config.get(...)`, which breaks the synchronous `config.get` contract.

Environment: node v24.18.0, config 5.0.0. (The same `defer(asyncFn)` pattern also leaves a Promise on 4.4.2 — on 4.x only the deprecated `asyncConfig` from `config/async` produced plain values.)

### What seems to be happening

Two things combine so that the resolved value never reaches the exported config object:

1. **`Util.resolveAsyncConfigs` only awaits; it never writes back.** In `lib/util.js` it collects `Promise` instances into an array and does `await Promise.all(promises)` — the properties on the config object it was given are never reassigned to the resolved values.
2. **The async `defer` write-back targets a stale object.** `deferConfig`'s async `resolve()` (`lib/defer.js`) does write the awaited value back via `Object.defineProperty(prop, property, ...)` — but `prop` is a node of the internal `Load.config`. The exported singleton is constructed afterwards by `new ConfigClass(load)` (`lib/config.mjs`), which copies the config, so the write-back lands on the internal pre-copy object and the exported config keeps the Promise. (`Util.resolveDeferredConfigs` also calls `defer.resolve()` without awaiting, so the copy is taken while the value is still pending.)

MMShep97:
Fixes #925

`Util.resolveAsyncConfigs` only awaited the promises it found — it never wrote the resolved values back. For async `defer` functions the write-back in `defer.js` lands on the internally loaded config, but the exported config is a *copy* of it (`Util.extendDeep` in the `ConfigClass` constructor copies the promise reference), so `config.get()` kept returning promises instead of values:

```js
const { Util } = require('config/lib/util');
const config = require('config');

Util.resolveAsyncConfigs(config).then(() => {
  config.get('fromDatabase'); // Promise { 'secret' } — expected 'secret'
});
```

Since v5 also removed `config/async.js` (whose `asyncConfig` did perform this write-back in 4.x), there was no working way left to resolve async values for synchronous `config.get()` access, despite the wiki documenting exactly this pattern ([Special features for JavaScript configuration files → Promises and async functions](https://github.com/node-config/node-config/wiki/Special-features-for-JavaScript-configuration-files#promises-and-async-functions)).

### Changes

- `Util.resolveAsyncConfigs` now replaces each promise it finds with its resolved value on the object it was given (using `Object.defineProperty`, since async `defer` write-backs can leave non-writable properties behind), restoring the removed `config/async.js` behavior.
- It resolves with the config object again, so the wiki's `resolveAsyncConfigs(config).then(config => ...)` chaining works.
- Unit tests for the write-back (plain promises, nesting, arrays, the copied-object case, rejections).
- `test/31-async-configs.js`: end-to-end regression test through the real exported config object and `config.get()`. The existing unit tests operate on plain objects, where the `defer.js` write-back masks the problem — that's why this slipped through.

No API changes; `npm run ci` (types + 331 tests) passes.

## Summary by CodeRabbit

- **New Features**
  - Asynchronous configuration values are now resolved throughout nested objects and arrays.
  - Resolved configuration is returned for immediate use.
  - Non-writable configuration properties are handled during resolution.
  - Rejected asynchronous values now propagate correctly.

- **Tests**
  - Added coverage for deferred values, nested configurations, arrays, composed values, and static settings.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -446,12 +446,16 @@ class Util {
   /**
    * Used to resolve configs that have async functions.
    *
+   * Replaces every promise-valued property on the given config object with its
+   * resolved value, so that subsequent (synchronous) `config.get` calls return
+   * plain values.
+   *
    * NOTE: Do not use `config.get` before executing this method, it will freeze the config object
    *
    * This replaces ./async.js, which is deprecated.
    *
-   * @param config
-   * @returns {Promise<void>}
+   * @param config the main config object, returned from require('config')
+   * @returns {Promise<typeof config>} resolves with the given config object once all promises are resolved
    */
   static async resolveAsyncConfigs(config) {
     const promises = [];
@@ -471,15 +475,23 @@ class Util {
         if (property.constructor === Object) {
           _iterate(property);
         } else if (property.constructor === Array) {
-          for (let entry of property) {
+          for (let index = 0; index < property.length; index++) {
+            const entry = property[index];
+
             if (entry instanceof Promise) {
-              promises.push(entry);
+              promises.push(entry.then((value) => {
+                Object.defineProperty(property, index, { value });
+              }));
             } else {
               _iterate(entry);
             }
           }
         } else if (property instanceof Promise) {
-          promises.push(property);
+          promises.push(property.then((value) => {
+            // defineProperty, because the property may be a non-writable value
+            // left behind by an async defer function (see ./defer.js)
+            Object.defineProperty(prop, name, { value });
+          }));
         }
         // else: Nothing to do. Keep the property how it is.
       }
@@ -488,6 +500,8 @@ class Util {
     _iterate(config);
 
     await Promise.all(promises);
+
+    return config;
   }
 
   /**
--- a/lib/util.js
+++ b/lib/util.js
@@ -455,7 +455,6 @@ class Util {
    * This replaces ./async.js, which is deprecated.
    *
    * @param config the main config object, returned from require('config')
-   * @returns {Promise<typeof config>} resolves with the given config object once all promises are resolved
    */
   static async resolveAsyncConfigs(config) {
     const promises = [];
@@ -500,8 +499,6 @@ class Util {
     _iterate(config);
 
     await Promise.all(promises);
-
-    return config;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


lib/util.js
```diff
@@ -488,6 +500,8 @@ class Util {
     _iterate(config);
 
     await Promise.all(promises);
+
+    return config;
```
jdmarshall: You’re not utilizing this anywhere. Do we really need the signature change?

jdmarshall:
@MMShep97 Can you rebase this on main? It seems I neglected to change the CI settings when renaming trunk and the tests aren't running on this at all.

jdmarshall (approved):

PR #929, opened 2026-08-20
types: restore full typedefs from DefinitelyTyped

### LINKED ISSUE #913: Removal of the `@types/config` package downgraded type safety
**Describe the bug**

Regarding the removal of the external [@types/config](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74605) package. While I'm fully in support of packages managing their types locally instead of through DefinitelyTyped, the decision did sadden me a bit since the new V4 types removed a lot of the type safety that I introduced in [#70288](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70288). 

I do acknowledge that the types I added in that PR were prone to TypeScript performance issues, particularly if any types were recursive, but I have since refined my approach to typing dot notation expressions and would really love to see that kind of type safety brought back to `node-config`. 

I'm more than happy to make a PR that restores these type definitions, but I wanted to first check to see if the maintainers are interested in this or if they no longer wish to have this kind of type safety.

TurtIeSocks:
This restores (with improvements) the type safety that the previous type definitions from DefinitelyTyped provided. 

[dot.paths](https://www.npmjs.com/package/dot.paths) is my own package and is purely type definitions, it does not increase bundle size and has been rigorously benchmarked with Fable 5 to make sure that it's as performant as possible. 

Fixes #913 

## Summary by CodeRabbit

* **New Features**
  * Added schema-aware TypeScript support for `config.get()`, including autocomplete and inferred value types for configured paths.
  * Supports nested configuration values and explicit generic type retrieval.
  * Preserved flexible string-based access when no schema is provided.

* **Tests**
  * Added coverage for typed configuration access, nested paths, property checks, and invalid paths or value types.
  * Expanded declaration compilation coverage for library types and schema-augmented configurations.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "5.0.1",
       "license": "MIT",
       "dependencies": {
+        "dot.paths": "^0.1.0",
         "json5": "^2.2.3"
       },
       "devDependencies": {
@@ -29,7 +30,7 @@
         "yaml": "^2.8.2"
       },
       "engines": {
-        "node": ">= 20.0.0"
+        "node": ">= 20.11.0"
       }
     },
     "node_modules/@bcoe/v8-coverage": {
@@ -942,6 +943,15 @@
       "license": "BSD-3-Clause",
       "engines": {
         "node": ">=0.3.1"
+      }
+    },
+    "node_modules/dot.paths": {
+      "version": "0.1.0",
+      "resolved": "https://registry.npmjs.org/dot.paths/-/dot.paths-0.1.0.tgz",
+      "integrity": "sha512-JhnbIboGe8ccS+ySWEBJ7+jxLQLl9uuWZeJfBeWK49iIZdXezFptOYjP+GX9MpRrJ86nw7qNHoigi7QCiEOn7Q==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=18"
       }
     },
     "node_modules/eachr": {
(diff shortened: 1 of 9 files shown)


package-lock.json
```diff
@@ -9,6 +9,7 @@
       "version": "5.0.1",
       "license": "MIT",
       "dependencies": {
+        "dot.paths": "^0.1.1",
```
jdmarshall, line 12: This is the part that's hanging me up. Does this need to be a dep instead of a devDep? Do we even need it at all?

TurtIeSocks: devDeps don't ship to users, so even if they did the augmentation, their config and its methods wouldn't be typed. 

I wrote it because I found myself continuously writing various versions of these types for multiple apps, that often had various performance issues or tradeoffs so I wanted to consolidate that effort into a highly refined type package that was reusable. So while I would appreciate the use of it in node-config, I made it MIT so its available to copy and paste in if you want to do that. 

Another option would be to declare it as an optional dependency and mention how to install it somewhere in the docs related to TypeScript, since non-ts users don't need it.

jdmarshall:
I don't want to absorb the supply chain of this one. I appreciate the Schema direction and some of the CI steps but is the rest of this necessary to get autocomplete to work?

TurtIeSocks:
> I don't want to absorb the supply chain of this one. I appreciate the Schema direction and some of the CI steps but is the rest of this necessary to get autocomplete to work?

Are we talking about the complexity of the types behind dot.paths? If so... yes? If you have a simpler solution for dynamically generating string literal types of the dot notation paths *and* their respective return types, without locking up the TypeScript parsing engine, I'm all ears. I'm not a TypeScript wiz but I have worked on this problem for quite a while.

PR #932, opened 2026-08-27
[#930] v5.0.2: Fix of ESM support in Jest

### LINKED ISSUE #930: Config 5.0 is not compatible with Jest
After upgrading node config from v4.4.2 to v5.0.0 my ESM Node v24 project fails to load it (at least inside Jest tests with enabled ESM support), because it asserts `config` is CJS, and throws `Unexpected import statement in CJS module`.

I see a few problems:

1. You don't declare `exports` in `package.json`, so by default Node uses `main` to resolve the package, thus `./lib/config.js`, which is CJS, and imports MJS module (but at least with that Jest/ESM support enabled it just assumes by default that modules loaded from CJS are all CJS as well).

2. If I add to your `package.json` the key `"exports": { "import": "./lib/config.mjs" }` my project correctly imports your `./lib/config.mjs` and considers it MJS, but then you still have other files with extensions `.js`, so my project considers them to be CJS (because to consider them MJS they either should have `.mjs` extensions, or your `package.json` should have `"type": "module"` key).

If I were you, I'd just make v5 MJS-only (_i.e._ having `"type": "module"` key, and all files in the distribution built as ESM) &mdash; most projects in the eco-system do it now, and especially in the case of `config` (which is very stable package) anybody who wants CJS version should have no real problems to stick to v4 &mdash; absolutely no need for backward CJS compatibility in v5.

birdofpreyru:
Fixes #930 

## Summary by CodeRabbit

* **New Features**
  * Added lazy, cached deferred configuration values.
  * Added ES module support for parsing JSON, YAML, TOML, XML, TypeScript, and other configuration formats.
  * Added package exports for improved compatibility with ESM and CommonJS applications.

* **Chores**
  * Updated module references and TypeScript declarations for ES module compatibility.
  * Expanded type resolution for supported module paths.


package.json
```diff
@@ -1,27 +1,30 @@
 {
   "name": "config",
-  "version": "5.0.1",
+  "version": "5.0.2",
```
jdmarshall: This fucks up GitHub’s release notes process. Don’t change version numbers if you’re not a maintainer.

jdmarshall:
Are you running the tests locally?

birdofpreyru:
> _Are you running the tests locally?_

Yeap, I overlooked at first that `npm test` does not run all the tests CI/CD is configured to run.

PR #933, opened 2026-09-03
Proposal: Logo for config.

jdmarshall:
I'm not wedded to this design, but it was more or less the first thing that popped into my head.

But the file size works better with OG link previews, which tend to scale or clip differently across platforms. 

## Summary by CodeRabbit

* **Documentation**
  * Added the project logo to the top of the README, including descriptive alternative text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+# <img src="media/logo.svg" alt="config"/>
+
 Configure your Node.js Applications
 ===================================
 
--- a/media/logo.svg
+++ b/media/logo.svg
@@ -1,0 +1,21 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="600" height="180">
+  <style>
+    text {
+      font-family: "Courier New", monospace;
+      font-size: 100px;
+      font-weight: 600;
+      fill: darkred;
+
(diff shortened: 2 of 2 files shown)


markstos:
The brackets are not specific to Node.js. Perhaps there's an option to more readily evoke the Node in node-config? 

alternate idea:  a Green hexagon to evoke Node.js logo with "C" or in the middle.

Node.js may not be able to use exactly the same green hex as Node.js does. That's a trademarked logo with its own trademark policy.

https://nodejs.org/en/about/branding

jdmarshall:
cnodejsHexnfig?

Actually the trademark guidelines specifically says NOT to change the color.

Here's a project that did a bit of what you're suggesting:  https://github.com/RafaelGSS/bench-node

markstos:
Config in a distinctive monospace font with a green hex for the O could work, as you suggest.

jdmarshall:
Tried another one. Unfortunately that's the end of my time for today.

markstos:
I endorse the new one. Simple and straightforward, like `node-config`.

markstos (approved):

